PR #73, opened 2012-05-15
Add line + column number to each log line

dhruvbird:
I think log lines should contain a line + column number so that debugging issues becomes easier. You no longer need to identify a log line with some sort of unique identifier if line numbers are present.

### DIFF
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -203,9 +203,15 @@ Logger.prototype.removeLevel = function() {
 };
 
 Logger.prototype.log = function() {
+    var location = new Error().stack.split('\n')[3].match(/:([0-9]+):([0-9]+)\)$/);
+    var category = this.category;
+    if (location) {
+        category += ':' + location[1] + ':' + location[2];
+    }
+
     var args = Array.prototype.slice.call(arguments)
-  , logLevel = args.shift()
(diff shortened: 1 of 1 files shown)


leegee:
This would be a really useful feature, on way or another. 
I don't mind it slowing-down my code when it is turned on, because it will only be turned on for either debugging or errors, and debuggin code is never deployed to production, and errors are (hopefully) rare.
Would it be possible to have a flag set when the conversion pattern is parsed, and then only gather and parse the Error.stack information if that flag has been set?

nomiddlename:
The stack trace has to be generated when the logging event is created - you'll get strange problems if you generate it in the layout used by the appender. My problem has always been working out how to add this expensive feature in without slowing things down for everyone (while still producing useful traces). Perhaps a good way of doing it would be to use the NODE_DEBUG environment variable to trigger generating traces in the logger?

leegee:
Why can't you use the Log4js configuration to turn it off/on, set a boolean if %l (etc) is found in the pattern layout when it is first met..?

ihmabreu:
You can always use the [log4js-extend](https://github.com/ww24/log4js-node-extend) while this important feature is not part of log4js.

nomiddlename:
Closing this one - it has been replaced by #580

PR #161, opened 2013-10-16
Make dateFile appender can keep the number of log files

venson999:


nomiddlename:
Totally missed this pull request - very sorry about that. I'm going to close it, because the code for file appenders has changed a lot since 2013, and this PR would only have worked for daily file rotations (not monthly, or hourly). Thanks for submitting it anyway.

PR #191, opened 2014-03-25
SMTP appender: use patternLayout for email subject

paypac:
This pull request changes the handling of the SMTP 'subject' option to use patternLayout instead of messagePassThroughLayout.

This means you can specify an appender like:

``` javascript
appender: {
    type: 'smtp',
    recipients: 'errors@company.com',
    subject: '%p on %h: %m'
}
```

Which will produce email subjects like: "ERROR on DEV-BOX03: foo is not a valid widget"

Apart from enabling pattern specifiers (%m etc), it works exactly as before:
- if subject is not specified, the subject will contain the first log message
- if a literal subject string is specified (eg, 'ERROR! please investgate...') it will be used verbatim

### DIFF
--- a/lib/appenders/smtp.js
+++ b/lib/appenders/smtp.js
@@ -15,7 +15,7 @@ var layouts = require("../layouts")
 */
 function smtpAppender(config, layout) {
 	layout = layout || layouts.basicLayout;
-	var subjectLayout = layouts.messagePassThroughLayout;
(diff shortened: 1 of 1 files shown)


nomiddlename:
I'm going to close this one. If you want to revisit this, please raise the PR against the [smtp appender repo](https://github.com/log4js-node/smtp).

PR #220, opened 2014-07-21
Fix #211. Try to create folders for logs.

### LINKED ISSUE #211: Error when log folder is not exist
For example, I config file appender's filename as `logs/log.log`, but if log folder is not exist, the error will occur.

PSyton:


nomiddlename:
This was fixed in streamroller. File appenders now create missing directories.

PR #229, opened 2014-08-14
added getBufferedLogger() function.

eurekaa:
Hi there.
This function should be useful when you need to log during async parallel operations, without having a mess in logs. 
For example when you walk asynchronously a directory and you want logs to be grouped by file.
This was my case and the getBufferedLogger() function did the trick to me.

It returns the same getLogger() object but messages are stored internally and sent to appenders only when you call the flush() method on it.

Hope it can be useful to others.
Thanks.
Stefano


nomiddlename:
Closing this one because this was replaced by another PR.

PR #240, opened 2014-08-29
Vivocha 0.6.20

lmalheiro:
Hi!
I had to add some features to log4js at my work and I thought those could be of interest. I added:
* subcategories: this allows to set a level for a major category then specify other levels for sub-categories. For example:
  "levels": {
    "cat": "WARN",
    "cat.sub1": "TRACE",
    "cat.sub1.sub12": "WARN",
    "cat.sub3": "INFO"
  }
If not specified, a sub-category inherits the level of its parent. Given the example configuration above, a logger with category "cat.sub4" would have level "WARN", inherited from category "cat". The appenders  also accept sub-categories, they will log the category and all sub-categories specified at their configurations. So, it is possible to create a logger only for category "cat.sub1.sub12".
* a level property at the file appender: this allows to restrict the levels that the file appender logs independent of the level configurations of the categories.
* a MARK level: this level is always logged and does not represent an error. Useful to send period marks to the log.
* a compression property at the rolling file appender: this activates compression of the rolled files. Unfortunately I can't tell if it works on other platforms than UNIX.

Let me know if it's worth to merge them,

Thanks,
Luis

### DIFF
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -6,7 +6,8 @@ var layouts = require('../layouts')
 , streams = require('../streams')
 , os = require('os')
 , eol = os.EOL || '\n'
-, openFiles = [];
+, openFiles = []
+, levels = require('../levels');
 
 //close open files on process exit.
 process.on('exit', function() {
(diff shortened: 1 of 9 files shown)


nomiddlename:
Build fails on node 0.8, but it might be time to abandon support for 0.8 anyway.

PR #257, opened 2014-11-02
Add support for time() and timeEnd() like in `console`

guss77:
The node console logger has a very useful timing function that allows the developer to easily generate operation timing logs to allow her to collect statistics on the run time performance of the application.

By replacing the console logger with log4js, this capability is lost. This pull request adds similar functionality with optional support for selecting the logging level of the time report.

Implementation is modeled after the log4javascript implementation, though it is new and (somewhat) different.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -99,7 +99,7 @@ Logger.prototype.time = function(name, level) {
 Logger.prototype.timeEnd = function(name) {
   timer = this.timers[name];
   if (!timer) return;
-  timer.report(logger);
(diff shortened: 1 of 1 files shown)


Glavin001:
Was this ever fixed and merged?

nomiddlename:
No, this never made it in. In the next version of log4js (2.x) I'm probably going to remove the console replacement from core anyway - it's a source of confusion and weird bugs. I might create a separate wrapper instead.

PR #261, opened 2014-12-01
added cluster identifier support

boljen:
I modified the clustered appender so it will add the original process identifier as well as an object containing cluster metadata.

I then modified the layouts.js file, first by rewriting the pid() function which will preferably use the loggingEvent.pid variable so that the log message will have the appropriate PID attached to it. I then implemented a new function, clusterInfo(), which allows you to write cluster-specific meta-data if available, with support for a custom pattern.

I'm not familiar with the unit-testing framework, hence no tests, but they should be pretty straight-forward.


nomiddlename:
Travis build error is unrelated to your change.

PR #266, opened 2014-12-17
Update smtp.js

sc2bigjoe:
added the ability for smtp appender to send message as html instead of plaintext. in your log4js.config file simply include "html": "true", to write out as html, otherwise it will send plaintext


nomiddlename:
thanks for this, seems reasonable.

PR #269, opened 2014-12-27
add options to reload function

Nekle:
if `cwd` is included in option, reload will not work


nomiddlename:
Thanks for this, looks sensible.

PR #277, opened 2015-03-04
Implemented shutdown function for SMTP appender

hasegawa-jun:


nomiddlename:
Thanks for this - I think it would be better to have a timeout, or a limited number of retries, otherwise this can keep the process open forever. For example, try for 5 seconds to send all messages and quit even if unsent messages remain.

hasegawa-jun:
Thank you for your feedback. Please take a look at inserted lines. I just added a timeout option which defaults to 5 seconds.

nomiddlename:
Released in 0.6.23

hasegawa-jun:
Thanks a lot!

PR #278, opened 2015-03-10
Custom connect tokens

cfogelberg:
The connect logger comes with a pre-defined family of tokens that can be used in its log string formats (e.g. `:url`, `:method`). This pull request extends the connect-logger to allow custom tokens (`arbitrary_string` or regexp) to be defined and used in the logger format.

Custom tokens are passed in and referred to in the `connectLogger` call like so:

```javascript
log4js.connectLogger(logger4js_logger, { format: ':method :custom', tokens: [{
  token: ':custom',
  replacement: 'this is not a very useful custom token'
}] });
```

If a custom token is the same as one of the default tokens (`:date`, `:response-time` etc) then the default token is overridden.

Some of the files had trailing whitespace which Sublime removed before I realised it - I've tried to separate those changes into separate commits but it makes for a slightly messy diff, sorry about that :/

### DIFF
--- a/lib/connect-logger.js
+++ b/lib/connect-logger.js
@@ -89,12 +89,12 @@ function getLogger(logger4js, options) {
 					if(res.statusCode >= 400) level = levels.ERROR;
 				}
 				if (thislogger.isLevelEnabled(level)) {
-          var default_tokens = assemble_tokens(req, res);
+          var combined_tokens = assemble_tokens(req, res, options.tokens || []);
 					if (typeof fmt === 'function') {
-						var line = fmt(req, res, function(str){ return format(str, default_tokens); });
+						var line = fmt(req, res, function(str){ return format(str, combined_tokens); });
 						if (line) thislogger.log(level, line);
 					} else {
-						thislogger.log(level, format(fmt, default_tokens));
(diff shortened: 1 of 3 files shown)


cfogelberg:
Quick note - the failing tests were failing in the original checkout as well so I didn't fix them, it looks like a date formatting thing relating to 24 hour clock versus 12 hour clock with AM and PM. Vows output:

```
    patternLayout 
      ✗ should default to "time logLevel loggerName - message" 
        »        
        actual expected 
         
        142:18:30 PM DEBUG multiple.levels.of.tests - this is a test 
         
         // layouts-test.js:13 
   
      ✗ %r should output time only 
        »        
        actual expected 
         
        142:18:30 PM 
         // layouts-test.js:13 
   
      ✗ %[%r%] should output colored time 
        »        
        actual expected 
         
        [36m1436m2:18:30 PM 
         // layouts-test.js:13 
  ✗ Broken » 1081 honored ∙ 3 broken (10.036s) 
  npm ERR! Test failed.  See above for more details.
```

nomiddlename:
Thanks for this. Tests that fail due to dates are very annoying, especially when they are intermittent - at some point I'll fix them up.

nomiddlename:
Released in 0.6.23

PR #280, opened 2015-03-14
enable chaining

a-x-:
Make possible write like this:

```
log4js.replaceConsole(logger.getLogger().setLevel('WARN'));
```

instead of:

```
log4js.replaceConsole((function(){var logger = getLogger(); logger.setLevel('WARN'); return logger;})());
```


nomiddlename:
Closing this one because I don't really think anyone wants to chain log messages, and there are too many conflicts now.

PR #282, opened 2015-03-20
Add appender-level timezone offset config

qbrandon:
Often enough remote servers have either diverging or irrelevant timezone settings (from the operation team point of view at least)
In order to display a relevant timestamp in the log, this PR implements an appender-level setting for timezone offset.
The impact is mostly on the prototype of layout that get an extra parameter for timezoneOffset.
The code assumes backward compatible behavior when such configuration is missing.
The following appenders have been modified:
* console
* dateFile
* file
* fileSync
* SMTP
Usage is as simple as adding a timezoneOffset to the appender configuration.
I am not too happy with some workarounds I had to use to get the unit tests to pass, but the original injection of getTimezoneOffset to override the default behavior was pretty hackish from the beginning making it difficult to fiddle with.
I guess a refactor of the timestamp handling to rely completely on UTC up until generating the final timestamp string could be a welcomed improvement.

### DIFF
--- a/lib/appenders/console.js
+++ b/lib/appenders/console.js
@@ -2,10 +2,10 @@
 var layouts = require('../layouts')
 , consoleLog = console.log.bind(console);
 
-function consoleAppender (layout) {
+function consoleAppender (layout, timezoneOffset) {
   layout = layout || layouts.colouredLayout;
   return function(loggingEvent) {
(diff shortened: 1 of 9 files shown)


nomiddlename:
Released in 0.6.23

qbrandon:
Excellent, thank you.

PR #286, opened 2015-04-01
Amazon SES Appender - using node-ses module

KrIsHnA-Revoked:
Quick run for https://github.com/aheckmann/node-ses

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Out of the box it supports the following features:
 * replacement of node's console.log functions (optional)
 * file appender, with log rolling based on file size
 * SMTP appender
+* Amazon SES appender (using node-ses)
 * GELF appender
 * hook.io appender
 * Loggly appender
--- a/lib/appenders/amazonses.js
+++ b/lib/appenders/amazonses.js
@@ -1,0 +1,119 @@
+"use strict";
+var layouts = require("../layouts"),
+    mailer = require("node-ses"),
+    os = require('os');
+
+/**
+ * SES Appender - sends mail using "node-ses" module
+ * 
+ * @param config appender configuration data
+ *    config.sendInterval time between log emails (in seconds), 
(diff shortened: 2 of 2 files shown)


KrIsHnA-Revoked:
This is my **first pull request** on github, Could anyone please tell me what happened **above**?

nomiddlename:
That's the Travis continuous integration build failing - it automatically builds and runs all the tests when someone submits a pull request. Your PR passes on node 0.10, but fails on node 0.8, but it looks like the failures are unrelated to your changes.

Could you explain why this appender is useful, please? I thought Amazon SES was an SMTP service, couldn't you just use the SMTP appender? Also, I would need tests that cover this appender as well.

KrIsHnA-Revoked:
Thanks for the info about Travis. :)

Amazon SMTP service can also be used., but, if someone want to be **quick and lightweight** with Amazon SES, want to implement only the **SendEmail** from the **Amazon Simple Mail Service** just like I thought, **node-ses** will be a good option. With **log4js-node** and **node-ses**, It took less than 30 minutes to write an error reporting service that sends all errors happened in the last 60 seconds altogether through an email only when an error occurs.

If the appender is available, after signing up with Amazon SES, we can set up an error reporting email service just under 5 minutes. So I thought this appender may be helpful if it is available with the **log4js-node** ( even it is pretty cool and easy to write new appender for **log4js-node**, Thanks for that ).

I'm not much aware of the standards and qualities required to get accepted my pull request. Also, I don't know, how can I write up a test that includes testing Amazon SES without compromising my keys. If you can point me to an example for this, that would be really helpful.

nomiddlename:
Hi, you should be able to mock out the node-ses calls, take a look at the smtp appender tests for an example.

nomiddlename:
Closing this one because it doesn't have any tests, and the functionality is already present in the SMTP appender.

PR #288, opened 2015-04-11
The smtp appender now works with the current version of nodemailer - fixes #287 

### LINKED ISSUE #287: smtp appender is not working with the current version of nodemailer
Here is the [migration guide](https://andrisreinman.com/nodemailer-v1-0/#migrationguide).

I patched this problem for my use and I hope, that I have time in the next few days to create a pull request.

Before doing this however, I have a minor question: Is there a way to use different log levels for different appenders? I.e. I want to log all levels into a file but only send mails on ```WARN```, ```ERROR``` and ```FATAL```.
If not, that would be great to have :-D

alawatthe:
I made as few as possible modifications. So every configuration which was working before should be working now. However, this is not compatible with the pre 1.0.0 releases of nodemailer.

### DIFF
--- a/lib/appenders/smtp.js
+++ b/lib/appenders/smtp.js
@@ -24,7 +24,7 @@ function smtpAppender(config, layout) {
 	function sendBuffer() {
     if (logEventBuffer.length > 0) {
 		
-      var transport = mailer.createTransport(config.transport, config[config.transport]);
+      var transport = mailer.createTransport(config.SMTP);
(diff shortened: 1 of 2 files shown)


alawatthe:
@nomiddlename Is there something, I should improve in this pull request, before it can be merged?

nomiddlename:
Thanks for this.

nomiddlename:
Released in 0.6.23

Issue #289, opened 2015-04-13
Hapi support


thebergamo:
Hello!

log4js not have any plugins to support hapi applications?

nomiddlename:
Closing this one because the Hapi support never made it in.

PR #290, opened 2015-04-14
Port Log4js for use in Hapi

thebergamo:
The tests only can running if hapi added on the package.json.

``` bash
npm install --save-dev hapi
```

The tests need improve, but its works.
#289

### DIFF
--- a/lib/hapi-logger.js
+++ b/lib/hapi-logger.js
@@ -103,6 +103,8 @@ function getLogger(server, options, next){
 	server.on('response', function(){
 		return next();
 	});
+
+	next();
 }
 
 /**
--- a/lib/hapi-logger.js
+++ b/lib/hapi-logger.js
@@ -1,0 +1,227 @@
+"use strict";
+
+var levels = require('./levels');
+var _ = require('underscore');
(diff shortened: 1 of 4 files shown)


nomiddlename:
Thanks for this - my only issue is that it looks like the version of hapi in package.json doesn't support node 0.8 (which log4js does), so the build fails. Any suggestions? Is there an older version of hapi that will still work?

thebergamo:
Well, I do some fix when I implemented in my projects. Because I don't understand very much about the Vowsjs I can't write better tests an any help is very encouraging and welcomed. 
About the issue, on tomorrow morning(GMT +3) I'll see this build more close and trying do it more compatible, but in version 8 of hapi some interfaces are broken, I will see how I can improve it =D

thebergamo:
Sorry about that, @nomiddlename support node v0.8 is a big Job, because I need use an oldest version of Hapi. I think maybe if you can update the CI node version for v0.10 and v0.12. (On version 0.12) the tests are broken.

What you mean?

nomiddlename:
log4js no longer supports versions of node.js below 4, so this might be worth revisiting now.

PR #294, opened 2015-04-27
Remove GELF flag when capturing custom fields

Skiggz:
While trying to set custom per-message fields via logging with the GELF appender, I noticed that logstash was dropping the message when the GELF flag was set. Still not 100% sure why, but a simple fix for me was to delete the GELF flag after it was checked (seen in this PR), and everything seemed to work happily.

TLDR sort of
* Logstash (which supports GELF) drops messages with the GELF flag present
* There's no particular need to keep this key besides telling the appender to set custom feilds

Hopefully I can merge this in, it seemed like overkill to mimic this appender for logstash with such a minor change.

### DIFF
--- a/lib/appenders/gelf.js
+++ b/lib/appenders/gelf.js
@@ -85,6 +85,8 @@ function gelfAppender (layout, host, port, hostname, facility) {
     var firstData = data[0];
     
     if (!firstData.GELF) return; // identify with GELF field defined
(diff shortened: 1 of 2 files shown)


mattdelliott:
LGTM

Skiggz:
@nomiddlename any chance someone could check this out?

nomiddlename:
Published to npm in version 0.6.25

PR #295, opened 2015-04-29
add protocol and hostname to output format for express

MaxXx1313:
You can use :protocol and :hostname in express format: 
log4js.connectLogger( log4js.getLogger(), { 
  level: 'auto', 
  format: ':remote-addr - :status - :hostname : :method :url' 
})


MaxXx1313:
Hey! All tests passes when i run it locally.

Issue #173, opened 2014-01-15
GELF appender problem


ticemochka:
Basically I have this error stacktrace
C:\workspace_frontend\chat-websocket-server\node_modules\log4js\lib\appenders\GELF.js:88
    if (!firstData.GELF) return; // identify with GELF field defined
                  ^
TypeError: Cannot read property 'GELF' of undefined
    at addCustomFields (C:\workspace_frontend\chat-websocket-server\node_modules\log4js\lib\appenders\GELF.js:88:19)
    at preparePacket (C:\workspace_frontend\chat-websocket-server\node_modules\log4js\lib\appenders\GELF.js:102:5)
    at Logger.<anonymous> (C:\workspace_frontend\chat-websocket-server\node_modules\log4js\lib\appenders\GELF.js:121:19)
    at Logger.EventEmitter.emit (events.js:117:20)
    at Logger.log (C:\workspace_frontend\chat-websocket-server\node_modules\log4js\lib\logger.js:54:8)
    at Logger.(anonymous function) [as info](C:workspace_frontendchat-websocket-servernode_moduleslog4jsliblogger.js:72:30)
    at Server.<anonymous> (C:\workspace_frontend\chat-websocket-server\lib\chatServer.js:36:16)
    at Server.g (events.js:175:14)
    at Server.EventEmitter.emit (events.js:117:20)
    at net.js:1052:10

It is happened when the value of logging message is undefined or null. I suppose this is an error.

sentient:
I noticed there are some changes coming up with https://github.com/nomiddlename/log4js-gelf
But I wonder if this issue has been fixed when a null value goes to the gelf appender.

I don't think we can use the log4js-gelf yet beause it requires log4js >= 0.7 

Examining all my log invocations on possible null is some work for just this appender.

darky:
Hey! It's a big trouble, that can destroy your app.

sentient:
I forked the repo and made a small change

https://github.com/sentient/log4js-node/commits/master

nomiddlename:
Fix pushed to NPM in version 2.3.0

PR #296, opened 2015-05-01
- check if data is available

sentient:
The GELF appender throws an error when data is null. This checks if the data is null.

var test;
logger.info(test);

### DIFF
--- a/lib/appenders/gelf.js
+++ b/lib/appenders/gelf.js
@@ -83,15 +83,16 @@ function gelfAppender (layout, host, port, hostname, facility) {
     var data = loggingEvent.data;
     if (!Array.isArray(data) || data.length === 0) return;
     var firstData = data[0];
-    
-    if (!firstData.GELF) return; // identify with GELF field defined
-    Object.keys(firstData).forEach(function(key) {
-      // skip _id field for graylog2, skip keys not starts with UNDERSCORE
-      if (key.match(/^_/) || key !== "_id") { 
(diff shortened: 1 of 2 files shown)


nomiddlename:
Could you add a test to cover this, please?

sentient:
ok. I will do that.

I am having some other thoughts as well. It would be nice to have some hooks into the logging framework. e.g. define a hook just before a message is send to the GELF appender (and allowing to define the context object). This very specific to GELF, but I had something similar with the File Appender. 

Have you considered hooks for customization? Or are there other mechanisme you would suggest before a message is send to a particular appender.

sentient:
@nomiddlename Do you have some documentation how I run the existing test framework?
Do you use grunt or gulp to kickstart this? 
I'm not familiar with .bob.json

sentient:
I think I got it
```
 > npm install bob -g

> npm install

 > bob test

> bob lint

....

```

nomiddlename:
If you just want to run the tests, "npm test" will do fine. Don't worry about the lint part.

sentient:
There was an issue that the 'should be compressed' is failing. I don't think that is related to the code change. Do all your test run fine at the moment?

nomiddlename:
This is a fix for issue #173 - I will look into this now that version 2 is out.

nomiddlename:
Thanks for the PR, I used your idea to fix the issue - but I couldn't merge the PR due to all the changes that have happened in version 2.x.

PR #301, opened 2015-06-04
Use x-forwarded-for if available for :remote-addr

jbruggem:
I just stumbled on this issue in my project and found out there was already a fix available :). 

Is this fix ok? Will you consider merging it in log4js-node?


nomiddlename:
Thanks.

nomiddlename:
Released to npm version 0.6.26

PR #306, opened 2015-06-16
hook to add cusotm layouts

gabe0x02:
Just a simple hook to add layoutMakers I then wrote a small repo to be able to log as json (https://github.com/gabe0x02/log4js-json-layout).  Let me know what you think.


nomiddlename:
Thanks for this - looks useful, and it has a test. Sorry it's taken me so long to review and merge.

PR #307, opened 2015-06-18
Change to end only writable file

kitakitabauer:
File that failed to open stream will result in an error in the fs.close() at exit event.
Please, check this pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "log4js",
-  "version": "0.6.25",
+  "version": "0.6.26",
   "description": "Port of Log4js to work with node.",
   "keywords": [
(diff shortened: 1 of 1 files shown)


nomiddlename:
Hi, I don't see in the [node docs for writable stream](https://nodejs.org/api/stream.html#stream_class_stream_writable) where the `writable` flag is defined. When does this get set? And if it is not documented, we shouldn't really be relying on it. I may have misunderstood your change. 

An alternative to using this writable flag would be to not add files to the list of open files if they failed to open a stream.

kitakitabauer:
Thank you! It will pull request again in the alternative!

PR #316, opened 2015-07-21
update dependencies

tgabi333:


nomiddlename:
Thanks for this - was there a reason for updating the dependencies? Do they not work on the latest node version or something?

PR #320, opened 2015-08-25
Added a way to register custom layout makers.

lud2k:
You register a custom maker with:
log4js.layouts.register('custom', function(config) {
  return function(event) { /* return formatted event here */ }
});


lud2k:
Not sure why the The Travis CI build failed. Is the latest version of log4js-node not passing?

nomiddlename:
Great minds think alike - I've just merged a PR that does the same thing (it won over yours because it had a test).

PR #329, opened 2015-10-12
Add missing category field

mercury2269:
Current logstashUDP appender doesn't send category of the loggingEvent which is very useful when filtering for logs.


nomiddlename:
Thanks for this.

nomiddlename:
Pull request #331 reverted your change, I think. On closer inspection it looks like the "type" field will contain the logger category unless you overwrite it in the config.logType property. Will this still work for you?

nomiddlename:
I've pushed pull request #331 to npm in version 0.6.28

PR #331, opened 2015-10-19
extra json data fields for logstash

jumpgh:
Extending log message with extra fields will allow to reduce parsing on logstash side.

### DIFF
--- a/lib/appenders/logstashUDP.js
+++ b/lib/appenders/logstashUDP.js
@@ -2,7 +2,6 @@
 var layouts = require('../layouts')
 , dgram = require('dgram')
 , util = require('util')
-, os = require('os')
 , _ = require('underscore');
 
 function log(loggingEvent) {
@@ -12,7 +11,7 @@ function log(loggingEvent) {
 function logstashUDP (config, layout) {
   var udp = dgram.createSocket('udp4');
   var type = config.logType ? config.logType : config.category;
-  layout = layout || layouts.colouredLayout;
+  layout = layout || layouts.dummyLayout;
   if(!config.fields) {
     config.fields = {};
   }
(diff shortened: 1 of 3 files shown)


nomiddlename:
Thanks for this. Could you add a test that covers this, please?

jumpgh:
I've updated logstash udp appender test

jumpgh:
Really impressed with your responsible attitude. 
So require('os') and commented out string are redundant and should be taken away. 

The same as calling layout. Logstash itself adds timestamp and host tags while type and log level should be placed into corresponding fields of json message structure. Available layouts serialize overal json message into escaped string making it totally useless without additional processing on logstash's side. For example, with layout you will get
``` 
{
     "@version" => "1",
     "@timestamp" => "2015-10-30T08:29:12.273Z",
     "type" => "nodejs",
     "host" => "10.0.3.1"
     "message" => "\e[32m[2015-10-30 11:29:12.273] [INFO] nodejs - \e[39mheartbeat { totalmem: 4035465216,\n  freemem: 221515776,\n}",
     "fields" => {
          "level" => "INFO",
          "totalmem" => 4035465216,
          "freemem" => 221515776
     }
}
```
while without layout you will get less messy message
```
{
      "@version" => "1",
      "@timestamp" => "2015-10-30T08:21:50.280Z",
       "type" => "nodejs",
       "message" => "heartbeat",
       "host" => "10.0.3.1"
        "fields" => {
             "level" => "INFO",
             "totalmem" => 4035465216,
             "freemem" => 136273920
    }
}
```
so plain text message looks much better for me, while you can add important values directly into fields

nomiddlename:
Thanks for explaining. I can see why you'd want structured JSON log messages. I'm only concerned that there might be users out there that are relying on the call to the layout function to add values in (pattern layout can be used to add process id, for example). I think that since by default it will use colouredLayout if none is provided (which I'm sure will mean that the messages have a lot of nonsense characters in them once they've been serialised), maybe we could change the default behaviour so that if no layout is provided in the config then none will be used (instead of using coloured layout). What do you think?

jumpgh:
So i guess, the best solution is to add FooLayout which would do nothing. Ok, I'll rewrite my code in that way.

jumpgh:
So, are there any unresolved issues?

nomiddlename:
I've merged this in, but I'm holding off on pushing to npm until I hear something from the owner of pull request #329 about the removal of the category field that he added to logstashUDP.

nomiddlename:
Never got a response about pull request #329, so I've pushed this to npm in version 0.6.28 anyway.

jumpgh:
thnx. upgrading my dependencies

PR #332, opened 2015-11-05
Patterns to describe caller

leegee:
Hi
Had a go at adding support for Log4j's `%l`, `%C`, and `%f` (line, class, filename).
This relies upon a change in `logger` to set once the call stack, but I have not yet found a good way to test this. 
Do you have any ideas?
The code is only an hour or so old, so needs testing.
TIA
Lee


leegee:
Not ready, sorry.

PR #334, opened 2015-11-18
Fix unitary test failing

gchauvet:
This patch contains 1 fix and small changes into .gitignore :
- Fixing a failed assertion who break integration process. The faulty assertion target a field that doesn't exists anymore.
- Reducing .gitignore by factoring regular expressions


nomiddlename:
Thanks for this.

PR #344, opened 2016-01-06
Remove the underscore dependency

hmalphettes:
Hi, I noticed that `package.json` was referring to `underscore` twice.
Given that there are only 2 places where it is used, I ventured into removing it all together.
I hope this helps.
Thanks for your attention!

### DIFF
--- a/lib/appenders/logstashUDP.js
+++ b/lib/appenders/logstashUDP.js
@@ -1,8 +1,7 @@
 "use strict";
 var layouts = require('../layouts')
 , dgram = require('dgram')
-, util = require('util')
-, _ = require('underscore');
+, util = require('util');
 
 function logstashUDP (config, layout) {
   var udp = dgram.createSocket('udp4');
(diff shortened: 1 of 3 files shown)


nomiddlename:
Cool - these changes make the code a little easier to read. I thought underscore was used in more places, but I guess not. Thanks for this.

hmalphettes:
Cheers!

I don't really have an agenda but if you you find it interesting to look at, I can give a shot at removing the handful of `async.series`, `async.each`, `async.eachSeries`, `async.whilst` with some recursive code.

nomiddlename:
If you've got the time, feel free. As long as it makes the code easier to understand, that's great - but I don't want to have to add too much complexity for the sake of removing a dependency.

shinnn:
@nomiddlename Can you publish a new version that include this change?

nomiddlename:
Pushed to npm in version 0.6.30

shinnn:
@nomiddlename Thanks!

PR #345, opened 2016-01-15
Replace async by native recursions

hmalphettes:
Hi, here is what it could look like if we were to replace async by native recursions.
I did this as a small exercise. I find it mostly more readable but it really is a matter of taste.
I also spotted that the shutdown test in logging-test.js is written as a synchronous test although it tests an asynchronous behaviour.
Thanks for your patience and attention!


nomiddlename:
Looks good. I think it does improve readability. Thanks a lot!

nomiddlename:
Pushed  to npm in version 0.6.30

PR #347, opened 2016-01-29
fix shutdown callbacks not being called

rlidwka:
Bug in https://github.com/nomiddlename/log4js-node/pull/345, `N.shutdown` currently never calls its callback in a few cases:

1. When no appenders with shutdown functions are loaded (for example, we're in a worker with only `clustered` appender loaded), the amount of shutdown functions is zero. So callback is never called 'cause it waits for a finish of at least one shutdown function.
2. When file appender is loaded, but not actually used, an amount of open files is zero. Same thing, callback is never called.

Same for dateFile apparently, although I didn't test that.

// cc: @hmalphettes

### DIFF
--- a/lib/appenders/dateFile.js
+++ b/lib/appenders/dateFile.js
@@ -65,6 +65,9 @@ function shutdown(cb) {
       cb(error);
     }
   };
+  if (!openFiles.length) {
+    return cb();
+  }
   openFiles.forEach(function(file) {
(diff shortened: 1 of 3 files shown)


hmalphettes:
Thanks. Sorry for the regression.

nomiddlename:
Thanks for this.

nomiddlename:
Published to npm in v0.6.31

PR #348, opened 2016-01-29
Feature/MailGunAppender

chewax:
## Mailgun Appender
Created an appender for using against mailgun API
The usage is very simple just add as any other appender. 
Im adding a simple usage example:

`appender: {
                  type: 'mailgun',
                  apikey: 'MAILGUN_API_KEY',
                  domain: 'MAILGUN_SENDING_DOMAIN',
                  from: 'sender@domain.com',
                  to: 'recepient@domain.com',
                  subject: 'Some Subject'
              }`

by the moment only supports one recipient at a time.
Im planning to extend that as soon as i have some time to spare.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "lib": "lib"
   },
   "dependencies": {
+    "mailgun-js": "^0.7.7",
     "readable-stream": "~1.0.2",
     "semver": "~4.3.3"
   },
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "lib": "lib"
   },
   "dependencies": {
-    "mailgun-js": "^0.7.7",
     "readable-stream": "~1.0.2",
     "semver": "~4.3.3"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "lib": "lib"
   },
   "dependencies": {
+    "mailgun-js": "^0.7.7",
```
nomiddlename: I don't put appender dependencies in to package.json - not all users of log4js will need mailgun, and they should not have to have it installed. People who want to use it will have to install it as part of their own dependencies (this is similar to the requirements for the smtp appender, for example).

chewax: Ok, should i take it out and add the requirement to the readme?
This is my first contribution to a project, if you dont mind guiding me towards what to do next that would be awesome

The error seems to be exactly that: the dependency missing from the project.

nomiddlename: Yep, remove it from the dependencies. Add a note to the README (or make a page in the wiki). Apart from that, everything looks good. Don't worry about the error - it's only for node 0.8, which I guess mailgun does not support. I'll look at updating the node versions that the travis build runs.

chewax:
Great! I have removed the dependency from package.json and added a page on the wiki.

nomiddlename:
Published to npm in v0.6.31

PR #349, opened 2016-02-01
feat(console.js): use native console.error for ERROR level logs and higher

feyy:
Only using `console.log` and color to distinguish levels isn't appropriate, especially when using with 'pm2'. The error logs will be writted in normal 'out' logs.

### DIFF
--- a/lib/appenders/console.js
+++ b/lib/appenders/console.js
@@ -1,11 +1,17 @@
 "use strict";
 var layouts = require('../layouts')
-, consoleLog = console.log.bind(console);
+, levels = require('../levels')
+, consoleLog = console.log.bind(console)
+, consoleError = console.error.bind(console);
 
 function consoleAppender (layout, timezoneOffset) {
   layout = layout || layouts.colouredLayout;
   return function(loggingEvent) {
-    consoleLog(layout(loggingEvent, timezoneOffset));
+    if (loggingEvent.level.level >= levels.ERROR.level) {
+      consoleError(layout(loggingEvent, timezoneOffset));
+    } else {
+      consoleLog(layout(loggingEvent, timezoneOffset));
+    }
   };
 }
 


nomiddlename:
Thanks for this. Does pm2 require your error logs to go to stderr? It sounds like you want to override the native console.* methods to get log4js style formatting, but you want error (and above) level messages to go to standard error instead of standard out - is this correct? There may be a better way to do that - I'm not sure that everyone who uses log4js will want the same thing. Perhaps we need a way to specify stdout and stderr as appenders separately, instead of just one "console" appender. What are your thoughts?

feyy:
Yes. I just want the error logs to go to standard error, your scheme is more suitable. Looking forward to your next release.

MichielDeMey:
I agree, we're currently using `supervisord` to monitor our logs and all logging levels appear in the `stdout` logs instead of actual errors going into the stderr logs.

lib/appenders/console.js
```diff
@@ -1,11 +1,17 @@
...
-, consoleLog = console.log.bind(console);
+, levels = require('../levels')
+, consoleLog = console.log.bind(console)
+, consoleError = console.error.bind(console);
 
 function consoleAppender (layout, timezoneOffset) {
   layout = layout || layouts.colouredLayout;
   return function(loggingEvent) {
-    consoleLog(layout(loggingEvent, timezoneOffset));
+    if (loggingEvent.level.level >= levels.ERROR.level) {
```
MichielDeMey: According to the node documentation, warnings are also logged to `stderr` since it's just an alias for `console.error`.
https://nodejs.org/api/console.html#console_console_warn_data

nomiddlename:
Does pull request #364 solve your problem?

MichielDeMey:
It does, but it feels weird that in order to print to `stderr` you would need the `stderrAppender ` but to write to `stdout` you would need the `consoleAppender`.

Feels like we should at least provide an alias `stdoutAppender` that aliases to `consoleAppender`.

Any thoughts around this?

nomiddlename:
Closing this because there are now stdout and stderr appenders.

PR #350, opened 2016-02-03
Update dependencies

shinnn:
npm v3 tries to dedupe the dependencies by default, and keeping dependencies up-to-date helps better deduplication.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,8 @@
     "lib": "lib"
   },
   "dependencies": {
-    "readable-stream": "~1.0.2",
-    "semver": "~4.3.3"
+    "readable-stream": "~2.0.5",
+    "semver": "~5.1.0"
   },
   "devDependencies": {
     "sandboxed-module": "0.1.3",


package.json
```diff
@@ -29,8 +29,8 @@
     "lib": "lib"
   },
   "dependencies": {
-    "readable-stream": "~1.0.2",
-    "semver": "~4.3.3"
+    "readable-stream": "~2.0.5",
```
shinnn: > readable-stream": "~1.0.2" -> "~2.0.5""

No breaking changes, but requires at least minor version bump.

package.json
```diff
@@ -29,8 +29,8 @@
...
+    "semver": "~5.1.0"
```
shinnn: > "semver": "~4.3.3" -> "~5.1.0"

Includes [a breaking change](https://github.com/npm/node-semver/commit/41b56fa573d1ef0d1718777a8a21ae1b57ebcd83), but it doesn't affect log4js.

nomiddlename:
I've updated the dependencies for the next release of log4js, thanks.

dangler:
@nomiddlename do you have an ETA on the next release? I want to use log4js and I'm having the semver dependency issues.

Thanks!

nomiddlename:
I just need to fix #414 and then I should be good to release.

PR #354, opened 2016-02-08
Improve logging performance

woollybogger:
Allow V8 to optimize the logging functions by **not [leaking the arguments object](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments)** and reducing the number of object accesses in the `toLevel` function.
This results in a performance improvement of about 11%.

##### Benchmark Results
before: 9,123 ops/sec ±1.46% (85 runs sampled)
after: 10,159 ops/sec ±1.97% (84 runs sampled)

##### Benchmark Code
```js
'use strict';

const Benchmark = require('benchmark');

log4js.configure({
  appenders: [
    {type: 'file', filename: 'log4jsbench.log', category: 'bench'},
  ],
  levels: {
    bench: 'debug',
  },
});

const logger = log4js.getLogger('bench');

const data = {a: 'thing', num: 2};

new Benchmark.Suite()
  .add('log', function() {
    logger.info('message', data);
    logger.trace('message', data);
    logger.log('info', 'message', data);
    logger.log('trace', 'message', data);
  })
  .on('cycle', function(event) {
    console.log(String(event.target));
  })
  .run();
```

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -13,20 +13,12 @@ function Level(level, levelStr) {
  * @type Log4js.Level
  */
 function toLevel(sArg, defaultLevel) {
-
   if (!sArg) {
     return defaultLevel;
   }
-
-  if (typeof sArg == "string") {
-    var s = sArg.toUpperCase();
-    if (module.exports[s]) {
-      return module.exports[s];
-    } else {
-      return defaultLevel;
-    }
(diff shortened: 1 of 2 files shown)


nomiddlename:
Thanks for this.

nomiddlename:
Published in npm version 0.6.32

PR #356, opened 2016-02-12
Hipchat connection

chrispotter:
@nomiddlename 

Here is a hipchat appender that I have added to log to a hipchat room using the [hipchat-client](https://github.com/germanrcuriel/hipchat-client) library.  

Please let me know what you think.  I based a lot of the tests and examples around mailgun and loggly, trying to keep with what seemed like best practices.  I would love to hear any suggestions you have.

### DIFF
--- a/lib/appenders/hipchat.js
+++ b/lib/appenders/hipchat.js
@@ -26,7 +26,7 @@ function hipchatAppender(_config, _layout) {
         var data = {
             room_id: _config.room_id,
             from: _config.from,
-            message: layout(loggingEvent, config.timezoneOffset),
+            message: layout(loggingEvent, _config.timezoneOffset),
             format: _config.format,
             color: colours[loggingEvent.level.toString()],
             notify: _config.notify
--- a/lib/appenders/hipchat.js
+++ b/lib/appenders/hipchat.js
@@ -52,5 +52,6 @@ function configure(_config) {
     return hipchatAppender(_config, layout);
 }
 
+exports.name      = 'hipchat';
 exports.appender = hipchatAppender;
 exports.configure = configure;
--- a/lib/appenders/hipchat.js
+++ b/lib/appenders/hipchat.js
@@ -1,0 +1,56 @@
+"use strict";
+var HipChatClient = require('hipchat-client');
+var layouts = require('../layouts');
+var layout;
+
+var hipchat, config;
+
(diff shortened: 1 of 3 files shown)


lib/appenders/hipchat.js
```diff
@@ -0,0 +1,57 @@
...
+            from: config.from,
+            message: layout(loggingEvent, config.timezoneOffset),
+            format: config.format,
+            color: colours[loggingEvent.level.toString()],
+            notify: config.notify
+        };
+
+        hipchat.api.rooms.message(data, function (err, res) {
+            if (err) { throw err; }
+            console.log(res);
```
nomiddlename: You probably don't want to log all the responses from hipchat, do you?

chrispotter: Good catch thanks!  Looks like I had a couple extraneous variable declarations too.

nomiddlename:
Thanks for this.

nomiddlename:
Published in npm version 0.6.32

luanped:
Do we need to manually npm install hipchat-client for this to work?, why is not as part of the package dependencies?

chrispotter:
Yes you need to include hipchat-client in your app package.json specified [here](https://github.com/nomiddlename/log4js-node/blob/master/examples/hipchat-appender.js#L2). For the same reason the mailgun client isn't included, I figured mailgun/hipchat isn't required to get the base log4js functionality working, rather just extensions/appenders as additions to the library. Therefore keeping the log4js install cleaner, it would be better to allow for individual inclusion of the libraries as needed. 

Also be sure that you're using API V 1.0 keys, which is what hipchat-client is based around.

luanped:
Thanks for your reply Chris. This isn't related to hipchat appender then, but however if without asking to use hipchat or mailgun or loggly, I'm getting all those cannot resolve module errors - does it mean something is configured wrongly on my side?

It looks to me that it tries to load every single appender in the lib/appenders folder, cause the moment I deleted everything but console the errors went away. 

Any insight will be appreciated!

chrispotter:
This might be a better question to be submitted with an issue, rather that on this closed pull_request, but it sounds like there is an issue with your configuration when initializing the log4js library.  You should post your code in an issue and see if anyone can help you there.  Sorry I can't be more help.

luanped:
No problem, thanks again :)

PR #357, opened 2016-03-03
Add new appender logglyTagged

justin808:
This appender allows adding tags to any call to the logger.

Documentation added:

Loggly Tagged Appender. Sends tagged logging events to Loggly using node-loggly.

The difference with the Loggly appender is that this appender will scan the msg from the
logging event, and pull out any argument of the shape `{ tags: [] }` so that it's possible
to add tags in a normal logging call. For example:

logger.info({ tags: ['my-tag-1', 'my-tag-2'] }, 'Some message', someObj, ...)

And then this appender will remove the tags param and append it to the config.tags.

### DIFF
--- a/lib/appenders/loggly.js
+++ b/lib/appenders/loggly.js
@@ -5,8 +5,46 @@ var layouts = require('../layouts')
 , passThrough = layouts.messagePassThroughLayout;
 
 
+function isAnyObject(value) {
+	return value != null && (typeof value === 'object' || typeof value === 'function');
+}
+
+function numKeys(o) {
+  var res = 0;
+  for (var k in o) {
+    if (o.hasOwnProperty(k)) res++;
+  }
+  return res;
+}
+
 /**
- * Loggly Appender. Sends logging events to Loggly using node-loggly 
+ * @param msg - array of args for logging.
+ * @returns { deTaggedMsg: [], additionalTags: [] }
+ */
+function processTags(msgListArgs) {
+  var msgList = (msgListArgs.length === 1 ? [msgListArgs[0]] : Array.apply(null, msgListArgs));
+
+  return msgList.reduce(function (accum, element, currentIndex, array) {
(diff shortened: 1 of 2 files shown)


ikrtolica:
This is super useful - Loggly recommends using tags to organize different services logging to their service - the output in Loggly looks something like:

image

justin808:
We'll add some tests the PR. There's some duplicated code with the logglyAppender.

We'd like to get some feedback before cleaning up any code. Thanks.

justin808:
I added tests.

There's a lot of duplication between the lib/appenders/logglyTagged.js and the lib/appenders/loggly.js files. 

I could reduce the duplication if deemed necessary.

Another option would be to replace the loggly.js file with the new one that supports tags.

nomiddlename:
I think it's probably a good idea to only have one loggly appender that supports tags - it'll get confusing otherwise.

justin808:
@nomiddlename Are you good with the API change? if the args include { tags: [] } then use those tags?

justin808:
I improved the tests and make the args check a bit more specific in that an object of the form { tags: [], others: []} is not considered to be an object that sets the tags.

justin808:
@nomiddlename I'll replace the existing loggly if you like and squash the commits.

nomiddlename:
Looks ok to me - the tags array object can appear anywhere in the log arguments, so it's not going to break things that don't use the loggly appender; and the existing behaviour of the loggly behaviour is preserved if you don't include a tag object. So go ahead and replace the existing one.

justin808:
@nomiddlename Done!

:shipit:

justin808:
@nomiddlename Any ETA of when this can be merged?

nomiddlename:
Sorry - got distracted by some shiny paper. I'll get on this now.

nomiddlename:
Published to npm in v0.6.33

PR #364, opened 2016-03-24
add a stderr appender

demmer:
Thus functions almost identically to the consoleAppender except that
instead of logging through `console.log` it uses `process.stderr.write`.

This is useful in cases where applications include their own console
output to stdout but want to have debug / error logs that go to stderr
so users can potentially redirect to separate out the logs from the output.


nomiddlename:
Thanks for this.

PR #365, opened 2016-03-25
Feature/use more appropriate console api functions

kayhadrin:
Proposing a way to use more appropriate console "log" functions so that we take advantage of the node or browser native console implementation.

i.e. 

- logger.debug will appear in blue in a browser
- logger.error will appear in red
- logger.warn will appear in orange in a browser,


nomiddlename:
Thanks for this, now that version 2.x is out I'll look at merging this.

PR #371, opened 2016-04-06
Shutdown for gelf appender

artem-karpenko:
Added a shutdown hook to gelf appender to be able to finish work w/o process.exit().


nomiddlename:
Thanks for this.

PR #373, opened 2016-04-11
Synchronous dateFile appender (With modes)

Smurf-IV:
Add new "thin" wrapper over DateRollingAppender, to allow 2 flavours of synchronous writing, 1 is (the default) via fsyncSync, the other via fsync via a config entry of `usefsync: true`.
Addition of unit tests based off of existing Date appender testing.

### DIFF
--- a/test/dateFileSyncAppender-test.js
+++ b/test/dateFileSyncAppender-test.js
@@ -153,7 +153,7 @@ vows.describe('../lib/appenders/dateFileSync').addBatch({
             {
               "category": "tests", 
               "type": "dateFileSync", 
-              "filename": "test/date-file-sync-test",
+              "filename": "test/date-file-test-sync",
               "pattern": "-from-MM-dd.log",
               "alwaysIncludePattern": true,
               "layout": { 


Smurf-IV:
This is not my test, so my code should still go in ?
test/streams/rollingFileStream-test.js
Does the fs emulation have to also include the fsyncSync methods ?

Smurf-IV:
Why is node 4.x LTS and Node 5.x not tested via **travis-cli** ?

Smurf-IV:
build 419.3 against 1345504 has this error:
`npm ERR! weird error 1` for the mailgun console write.
This build system does not inspire confidence !
Anyway, Please accept the changes.

Smurf-IV:
[Created an explanation page here:](https://github.com/nomiddlename/log4js-node/wiki/Date-rolling-file-appender---with-synchronous-file-output-modes)

nomiddlename:
I'm a bit confused as to the purpose of this. Calling `fsync` after every write sounds expensive. I would have thought a synchronous appender would just use `writeSync`. Perhaps it's a naming problem - this seems to be more of a "guaranteed write" appender than a synchronous appender, since all the file operations are still asynchronous. 

If you agree, then this functionality should be moved into BaseRollingFileStream (with a "fsyncAfterWrite" option), so that it can then be used by both rolling file streams, not just the date file roller. What do you think?

nomiddlename:
(Oh, and yes - I agree that I should update the travis config to use some more up-to-date node versions, thanks for reminding me)

Smurf-IV:
fsyncSync is the default, and yes it would be "partially blocking", all Sync calls in Node are "expensive", but still hand-off to the underlying OS to perform the write whilst "they" just wait for the call to return.

Adding into the base class would not allow for the "2 types" of sync call to be selected without an "Enum / text" check requirement. (Not impossible just not as simple)

Also if moving to the base class then the FileSync should also be refactored to be a base class override.

Thoughts ?

nomiddlename:
Closing this because there have been a lot of changes to the file appenders since this was first proposed.

emouawad:
Any updates on this? Has this been merged yet? if not @Smurf-IV  do you have a working Lib?

Issue #122, opened 2013-03-11
Logs not flushing to file on process.exit


stormherz:
Problem, described in pull request https://github.com/nomiddlename/log4js-node/pull/65, logging something to file appender and ending process right after results in missing logs.

despairblue:
@nomiddlename does this also flush to loggly? I was trying to find a way to flush messages to loggly before exiting, and it seems the only viable solution is `setTimeout`.

nomiddlename:
The Loggly appender doesn't register a shutdown function, which is probably the issue there.

despairblue:
@nomiddlename would you welcome a PR?

PR #378, opened 2016-04-19
Feature/loggly shutdown

despairblue:
Follow up to #122.

I’d like to add a test for this, but it seems that appenders that are registered with `addAppender` cannot register a shutdown handler. Is this on purpose, or has this simply be forgotten when adding the shutdown handlers?

I’d be happy about any hints. Thanks.

P.S.: I also tried to adhere to the .jshintrc, though lots of files in this repo don’t. Is this file up to date? And could we add jshint as a dev dep and make it a pretest or posttest script?

### DIFF
--- a/lib/appenders/loggly.js
+++ b/lib/appenders/loggly.js
@@ -2,8 +2,9 @@
 var layouts = require('../layouts')
 , loggly = require('loggly')
 , os = require('os')
-, passThrough = layouts.messagePassThroughLayout;
-
+, passThrough = layouts.messagePassThroughLayout
+, openRequests = 0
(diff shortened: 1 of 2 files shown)


nomiddlename:
Yep, looks like you're right about addAppender bypassing the shutdown function registration. I don't think this is on purpose, most likely it was just forgotten. Maybe we could add an extra argument to addAppender for passing that in?

Thanks for trying to follow the jshint rules - I know I've been lax in making sure everything applies. Definitely a good idea to add it as part of the build.

nomiddlename:

PR #387, opened 2016-05-10
logFaces appender

sparklton:
This addition contains logFaces UDP appender. It serializes log events into compact JSON string and sends them to logFaces server over UDP. Configuration examples are included. Tests are added. I hope it will be a useful addition to this project.


sparklton:
Some of the tests are failing on older versions of node.js like v0.12 and v0.10. Anything needs to be done on my part to make it pass?

sparklton:
OK, I've figured it out eventually. Older versions of node can't require process.js which I used in this appender. So, it can be safely ignored when it fails to load, the appender will do without it. Looking forward for this pull request to be finalized.

nomiddlename:
Thanks for this.

nomiddlename:
Released in log4js@0.6.37

PR #395, opened 2016-06-20
Fixes #393 - catch error when sending UDP packet to GELF server

### LINKED ISSUE #393: GELF appender throws unhandled error if DNS lookup fails
```javascript 
var log4js = require('log4js');
log4js.configure({appenders:[{host: 'IDontExist.example.com', port: 12201, type: 'gelf'}]});
var l=log4js.getLogger();
l.info('test');
```

Returns: 
```
events.js:160
      ^

Error: getaddrinfo ENOTFOUND IDontExist.example.com
    at errnoException (dns.js:28:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:79:26)
```

This can be solved in 2 ways:
1. Listen on client errors
2. pass a callback to client.send()

dcolens:


nomiddlename:
Good catch - could you add a test that covers this, please?

dcolens:
I looked at the tests when I made the patch, but my vows experience is none, and I'm not too sure how to write one. Even-though it is trivial to repro :/

nomiddlename:
Yeah, fair enough - vows is a bit dodgy.

PR #396, opened 2016-06-21
make HTTP response body available for formatting function

akrasov:
in some cases we want to log HTTP response body,  so we include a response body (Buffer) to the res object before formatting function is called.


nomiddlename:
Is there a way to make this optional? - keeping a copy of the response body around is not something everyone will want.

nomiddlename:
And you've also included a change to the GELF appender, was that an accident?

nomiddlename:
Closing this because it should be optional, and the PR also includes some other changes.

PR #398, opened 2016-07-11
add req.url so that connect works with restify

abrarahmedbcg:
restify uses req.url to denote the invoked url path for http, this change allows to make the logger connect work with restify out of the box


nomiddlename:
Thanks for this.

PR #399, opened 2016-07-16
ES6 refactor

e-cloud:
### Reason

Sooner or later, [the maintenance of old version of Node.js would be no longer supported](https://github.com/nodejs/LTS)

Therefore, evolving the code to get get fundamental support and keep update with the community is necessary.

I believe, with more clear structure, the maintenance and the evolution of this library would be much more easier.
### Changes

Basically, the refactor is of no logical changes except some constructor initializing order of classes  inside streams changed. One evidence is that they pass the nearly unchanged tests.

Besides, the fundamental facilities are also improved for better code management.

---

> NOTE: Now there is some conflict with `master` branch. If @nomiddlename you are willing to accept these changes, the conflict will be gone soon.
> 
> NOTE2: There are test related dependencies -- `sandbox-module` and `vows`. They can't be easily updated. `Vows` is even because of its internal bug(s).


nomiddlename:
I'm happy with removing support for v0.10 and v0.12, but I'd like to support v4 for a while. I don't think there's much you've used that won't work in v4 though, as far as I can tell - am I right?

e-cloud:
Actually, i don't know what are not supported in v4 clearly.
YES, I would have a test on v4 soon.

nomiddlename:
[Here's a handy table of ES6 support in different versions of node](http://node.green)

e-cloud:
by some fallback, this PR works with Node.js latest v4(not sure with the old v4).

It's a bit pity that fallback the rest parameter syntax :sweat_smile:

e-cloud:
hey, @nomiddlename. I just got time to sync the branch with the master. I managed to upgrade some dependencies and adjust the code.

However, here it reveal two problems of the (test) code. You can see the details in the latest commit of this PR. We may discuss about it once you have a review.

nomiddlename:
Wow. Thanks for this - looks like a lot of work has gone into it. I'll review over the next couple of days, but I'm thinking this would form a 2.x release.

nomiddlename:
Looks like that logging-test.js where it expects the invalid config to throw an exception can't have worked for a while. You're right about it exposing a bug in the library. I'll add some validation of the config. I'm going to merge this into master, but I won't push the changes to npm until I've changed a few other things in the lib. Thanks again for doing this, very helpful.

e-cloud:
sure, unstable code shouldn't be published.

PR #400, opened 2016-07-27
added categoryname to logstash appender

segler-alex:


nomiddlename:
Seems simple enough :) - I don't use logstash, so I'm going to assume this is correct.

nomiddlename:
Released in version 1.0.0

Issue #338, opened 2015-11-30
datefile appender memory leak


xinyue503:
My configuration：
log4js.configure({
    appenders: [
        {
            type: 'console',
            category: "console"
        }, //控制台输出
        {
            type: "dateFile",
            filename: serconf.logPath + 'log4jstest.log',
            pattern: "_yyyy-MM-dd",
            alwaysIncludePattern: false,
            category: 'dateFileLog'
        }//日期文件格式
    ],
    replaceConsole: true,   //INFO,DEBUG,ERROR,WARN
    levels:{
        dateFileLog: 'DEBUG'
    }
});

the program Use jmeter to conduct stress tests 1000 threads, running for an hour, hang up

How should handle?

royalrover:
I have the same question.
I analyse the heapdump, and find lots of "writeableState" & "DateRollingFileStream" Object. These Objects costs 90% of mem。

nomiddlename:
I've tried reproducing this with the following script:
```javascript
var log4js = require('../lib/log4js');
log4js.configure({
  appenders: [
    {
      type: 'console',
      category: "console"
    }, //控制台输出
    {
      type: "dateFile",
      filename: 'log4jstest.log',
      pattern: "_yyyy-MM-dd",
      alwaysIncludePattern: false,
      category: 'dateFileLog'
    }//日期文件格式
  ],
  replaceConsole: true, //INFO,DEBUG,ERROR,WARN
  levels:{
    dateFileLog: 'DEBUG'
  }
});
var logger = log4js.getLogger('dateFileLog');
// log every 30ms
var tid = setInterval(function () {
  logger.debug("cheese and crackers");
  console.log('some other logging thing.');
}, 10);

// stop after 30s
setTimeout(function() {
  clearInterval(tid);
}, 30000);
```
The only memory leak I see when profiling is the known issue with the console appender. If you replace the console appender with `stdout` appender, then the memory allocation is pretty flat. If you can give me any more information, or even better some code that reproduces the issue, that would be great.

nomiddlename:
Closing this issue because I can't reproduce. Happy to look at it again in future, if there is a way of reproducing the problem.

davidvivi:
> Closing this issue because I can't reproduce. Happy to look at it again in future, if there is a way of reproducing the problem.
I reproduce it,
![image](https://user-images.githubusercontent.com/19609057/98797842-13996f80-2448-11eb-92a0-74700b92f8da.png)

[...]

PR #401, opened 2016-08-12
logFaces append via HTTP

sparklton:
logFaces appender is updated with capability to send log events over HTTP for being useful on client side as well. UDP sending is not changed.


nomiddlename:
Hi, thanks for this - apologies for taking so long to get around to it. A couple of issues - there are no tests that cover the HTTP part, and the HTTP part seems to rely on axios which isn't mentioned anywhere. How will users know they need to install that as well? It also looks like you've changed the format of the UDP messages (different fields).

sparklton:
Hi Gareth, I didn't think there was need to test actual network behavior. Both cases (UDP and HTTP) have network calls stabbed for simplicity. As long as appender works as configured, the test coverage should be sufficient. I think. Am I wrong? I thought that since network dependencies for UDP and HTTP are carried out by modules which have sufficient test coverage, there is no need to test them here. You are right, I should have added some explanation about axios dependency. This appender can be used in server side (node.js) and client side (browserified) where axios was the best choice for doing HTTP calls, we can't do UDP from browsers, this is why HTTP option was added. I will add some "user manual" shortly.

nomiddlename:
You're right - my fault, I misread the test code. Sorry about that.

sparklton:
No problem at all, I've just added some clarifications and links to dependencies.

nomiddlename:
Thanks for this. I'll include it in the 1.1 release (I want to fix #338 before I release that though)

nomiddlename:

PR #403, opened 2016-08-28
Add close-on-SIGHUP behavior for logrotate.

### LINKED ISSUE #343: Close and reopen logs on signal
Systems using `logrotate` need to be able to tell a logging process to close and reopen its logs after it has rotated them, so the process doesn't keep logging to the old inode. Nginx for instance implements this with the -USR1 signal. The file appender in log4js should have a method to close and reopen files, either called from user code or registered as a signal handler.

jsha:
Fixes #343


nomiddlename:
Thanks for this - I finally got around to merging it and writing a test.

jsha:
Excellent, thanks for the merge (and the test)!

nomiddlename:
Fixed in version 1.0.0

GiteshKhanna:
So which shell command should we be writing in the postrotate of logrotate?

nomiddlename:
You probably want something like `kill -HUP <pid of your app process>`

GiteshKhanna:
> You probably want something like `kill -HUP <pid of your app process>`

So I believe, since I have a java springboot application running slf4j, should I be killing the springboot application? Won't it effect any other services as well?

nomiddlename:
This is an issue about a log library for node.js applications, nothing to do with java, sorry.

Issue #315, opened 2015-07-20
Allow log4js to do milliseconds.


pereran:
changed date_format.js:

exports.ISO8601_WITH_TZ_OFFSET_FORMAT = "yyyy-MM-ddThh:mm:ss.SSSO";

thu2004:
Just pushed a pull request: https://github.com/nomiddlename/log4js-node/pull/420

PR #420, opened 2016-11-10
fix for #315 - Include milliseconds in format ISO8601_WITH_TZ_OFFSET

thu2004:
Without milliseconds, the output of the log when sent to log collector, such as Logstash or Kibana, are not useful. Because Kibana uses the timestamp to sort the log, then the order of the log not always display correctly within one second.


nomiddlename:
Thanks for this, I'll include it in the next release.

nomiddlename:

PR #421, opened 2016-11-10
Improve core logging performance

nwoltman:
Allow V8 to optimize functions by not [leaking the arguments object](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments).
This results in a performance improvement of about 20%.

#### Benchmark Results

before: 28,566 ops/sec ±2.27% (82 runs sampled)
after: 34,305 ops/sec ±1.96% (87 runs sampled)
Improvement: ~20%

#### Benchmark Code

``` js
'use strict';

const Benchmark = require('benchmark');

log4js.configure({
  appenders: [
    {type: 'file', filename: 'log4jsbench.log', category: 'bench'},
  ],
  levels: {
    bench: 'debug',
  },
});

const logger = log4js.getLogger('bench');

const data = {a: 'thing', num: 2};

new Benchmark.Suite()
  .add('log', function() {
    logger.info('message', data);
  })
  .on('cycle', function(event) {
    console.log(String(event.target));
  })
  .run();
```

### DIFF
--- a/lib/layouts.js
+++ b/lib/layouts.js
@@ -33,7 +33,7 @@ function wrapErrorsWithInspect(items) {
         if (semver.satisfies(process.version, '>=6')) {
           return util.format(item);
         } else {
-          return util.format(item) + '\n' + item.stack;           
+          return util.format(item) + '\n' + item.stack;
         }
       } };
     } else {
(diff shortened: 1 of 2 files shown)


nomiddlename:
Awesome - thanks for this, it will be included in the next release (I'm just waiting on some feedback about #338 before pushing out the next one).

nomiddlename:

PR #424, opened 2016-12-06
Added support for passing a function/closure to the logger

cadement:
Added support for passing a function/closure to the logger as a message instead of a raw string.  This allows log messages to be lazily evaluated only when actually emitted (which can yield substantial performance benefits).  This was already possible using logging level-based conditional logging, but the syntax was much messier:

    if (logger.level.level < 5000) logger.trace(JSON.stringify(myBigObject));
  vs
    logger.trace(() => JSON.stringify(myBigObject));

### DIFF
--- a/lib/layouts.js
+++ b/lib/layouts.js
@@ -51,6 +51,9 @@ function formatLogData(logData) {
       data[i] = arguments[i];
     }
   }
+  
+  data = data.map(function(elem) { return (typeof elem == 'function') ? elem() : elem; });
+  
   return util.format.apply(util, wrapErrorsWithInspect(data));
 }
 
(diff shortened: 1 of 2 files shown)


nomiddlename:
I can see this being useful, so thanks for the PR. My only concern is that there could be someone out there that's logging a function somehow, and this would now execute that function. Would you be able to add a config parameter to turn this feature on and off - with "off" being the default. A boolean at the top level of config would be fine.

nomiddlename:
Leaving this open because it's still a good idea, but it'll need to some work to port to tap tests and allow for turning it off and on.

ericnewton76:
IMO this is a great feature and its more natural.  i would recommend a major version bump on this, and allow the config default to be true, noting that in a v2->v3 upgrade guide

PR #432, opened 2017-01-17
feat(LogstashUDP): Added a way for fields to be added without dot.

franciszabala:
I was trying to create a more consistent way of creating fields by removing the "field" and the dot for LogstashUDP. For some reason, Logstash 5.1 mutate does not process any fields with a "." in it's field.


nomiddlename:
Thanks for this, much appreciated. I'll include it in the next major release of log4js.

Issue #438, opened 2017-02-14
loadConfiguration from file when doesnt exist causes error


ericnewton76:
In the vein of existing log4j and log4net, an attempt to load a configuration file that doesnt exist shouldnt error.

This is so logging could be turned on later, after distribution of the library using log4js.

I'm thinking to do a PR to do this.

I'll just disable error when configuration file doesnt exist, and eventually another PR to support autoreload.

nomiddlename:
Sorry @ericnewton76  - I've been away on holiday, so haven't been able to look at this. I've already done a lot of work on configuration validation, in the `version-2` branch - you might want to take a look at that and see what you think.

nomiddlename:
Closing this - I'm of the opinion that log4js *should* complain about missing config. If this is for a library that wants to support log4js for users, then take a look at using [log4js-api](https://www.npmjs.com/package/@log4js-node/log4js-api)

PR #439, opened 2017-02-14
updated loadConfiguration to not raise exceptions.

ericnewton76:

addresses Issue #438


ericnewton76:
somehow managed to push without eslint running, and the tap tests wont run on my machine (windows issues?)

PR #440, opened 2017-02-14
fix: loadConfiguration errors when file doesnt exist or not parseable

ericnewton76:

addresses Issue #438


ericnewton76:
btw, the npm tests dont succeed on my Windows machine.

also, need to update package json for Windows machines to call tap properly:

```
  "scripts": {
    //...//
    "test": "tap \"test/tap/**/*.js\"",
    //...//
  }
```

PR #441, opened 2017-02-15
Feature/load configuration file less errors

ericnewton76:

addresses Issue #438

### DIFF
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -294,7 +294,21 @@ const configState = {};
 
 function loadConfigurationFile(filename) {
   if (filename) {
-    return JSON.parse(fs.readFileSync(filename, 'utf8'));
+    
+    let configStr;
+    try {
+      configStr = fs.readFileSync(filename, 'utf8');
+    } catch (ex) {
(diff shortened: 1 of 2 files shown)


nomiddlename:
Thanks for your work (and it's a really good quality PR), but I'm not going to merge this change - I think it should be the application's responsibility to handle missing config. The app can catch the error thrown by log4js if it knows the config file might not be there.

ericnewton76:
Tbh I would go with silent approach for errors in the logging framework unless configuration (usually during a special logging system debug session) at runtime. 

See NLog for same philosophy, meaning that yes logging is important but shouldn’t (ideally) introduce additional runtime headaches. 

They have support for a few environment variables to configure this behavior which I’ve used many tiles to evaluate and fix logging framework related problems

Would you be open to reopening this?

nomiddlename:
Issue #575 was raised about how to handle I/O errors in appenders, and it might be worth combining these two to come up with a unified approach to handling errors caused by log4js - @ramicohen303 did you have something in mind? Perhaps you and @ericnewton76 could work together or bounce ideas off each other?

PR #442, opened 2017-02-15
Fix/npm scripts for windows

ericnewton76:
fix to get npm scripts to run in windows environment.  single quote seems to cause things to fail whereas using doublequotes works.


ericnewton76:
sorry branched from my other work, recreating pull request from tip of nomiddleman/log4js/master

PR #443, opened 2017-02-15
fix: npm scripts in packagejson changed to use doublequote

ericnewton76:
fix for npm scripts in windows environments.

single quotes seems to cause problems. 

for instance: `npm test` run on windows environment tries to run `tap 'test/tap/**/*.js'` using single quotes but node in windows seems to interpret the single quotes different.

change these to double quotes and it works fine: `tap "test/tap/**/*.js"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,15 +22,15 @@
     "node": ">=4.0"
   },
   "scripts": {
-    "clean": "find test -type f ! -name '*.json' ! -name '*.js' ! -name '.eslintrc' -delete && rm *.log",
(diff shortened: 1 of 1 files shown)


ericnewton76:
i'm completely at a loss as to why changing the single quotes to double is causing this test to fail... i feel like these are unrelated:
https://travis-ci.org/nomiddlename/log4js-node/jobs/201933671#L907

ericnewton76:
ugh just keep this on hold i'm researching the problem

nomiddlename:
Closing this, on the assumption that a newer NPM fixes the problem.

PR #449, opened 2017-03-02
Bump streamroller to 0.4.0

hansonw:
Would it be possible to do a patch release (1.1.1)? We rely on this over at https://github.com/facebook/nuclide for logging and the unhandled exceptions are causing server crashes :(


nomiddlename:
I've pushed version v1.1.1 to npm now, with the updated streamroller dep. Let me know that it works for you.

hansonw:
Awesome, thank you so much!

Issue #385, opened 2016-05-04
fileAppender write after end, 


leegee:
Log4js-node: `"version": "0.6.35"`

[package.json.txt](https://github.com/nomiddlename/log4js-node/files/248730/package.json.txt)

Circumstance: with a large log file, or any empty log file.

Configuration:

``` json
{
    "appenders": [
        {
            "type": "categoryFilter",
            "exclude": [],
            "appender": {
                "type": "console",
                "layout": {
                    "type": "pattern",
                    "pattern": "%[%-10c %-5p%] %m"
                }
            }
        },
        {
          "type": "file",
          "filename": "temp.log",
          "maxLogSize": 20480,
          "layout": {
             "type": "pattern",
             "pattern": "%[%-10c %-5p%] %m"
          }
        }
    ],
    "replaceConsole": true
}
```

Output (repeated many, many times):

```
[31mconsole    ERROR[39m log4js.fileAppender - Writing to file temp.log, error happened  [Error: write after end]
Error: write after end
    at writeAfterEnd (_stream_writable.js:159:12)
    at RollingFileStream.Writable.write (_stream_writable.js:204:5)
    at Logger.<anonymous> (D:\src\thin-monitor2\node_modules\log4js\lib\appenders\file.js:68:13)
    at emitOne (events.js:82:20)
    at Logger.emit (events.js:169:7)
    at Logger._log (D:\src\thin-monitor2\node_modules\log4js\lib\logger.js:91:8)
    at Logger.(anonymous function) [as info] (D:\src\thin-monitor2\node_modules\log4js\lib\logger.js:83:14)
    at Console.log (D:\src\thin-monitor2\node_modules\log4js\lib\log4js.js:359:10)
    at Object.puts (C:\Users\lee\AppData\Roaming\npm\node_modules\protractor\built\logger.js:24:17)
    at log_ (C:\Users\lee\AppData\Roaming\npm\node_modules\protractor\built\launcher.js:24:12)
```

Followed by:

```
RangeError: Maximum call stack size exceeded
    at Function.getOwnPropertyNames (native)
    at cloneCallSite (C:\Users\lee\AppData\Roaming\npm\node_modules\protractor\node_modules\source-map-support\source-map-support.js:301:10)
    at wrapCallSite (C:\Users\lee\AppData\Roaming\npm\node_modules\protractor\node_modules\source-map-support\source-map-support.js:328:13)
[...]

PR #454, opened 2017-03-08
Properly flush file appenders on shutdown

### LINKED ISSUE #270: Datefile and file appenders add blank lines on shutdown
Hello

When I shutdown log4js using log4js.shutdown, the shutdown function gets called in the configured dateFile appender (this function also would get called in the regular non-sync file appender).

Shutdown tests the stream by attempting to write a new line. I think this should be changed to instead write an empty string ('') such that the logs don't end up with an extra white line written to file.

Thoughts? 

```
function shutdown(cb) {
  async.each(openFiles, function(file, done) {
    if (!file.write(eol, "utf-8")) { /// -> change this to !file.write ('', "utf-8")
      file.once('drain', function() {
        file.end(done);
      });
    } else {
      file.end(done);
    }
  }, cb);
}
```

hansonw:
The current file appender `shutdown` functions don't actually work properly.
`write` is fully asynchronous whether it returns true or false, so we can't just `end` the stream immediately - this causes a "write after end" error.

I think the right thing to do here is to write an empty string and wait for the callback to trigger. Since writes are sequential and logging gets disabled prior to shutdown, this should ensure that any earlier writes have already occurred.

In addition, I don't think we actually need the draining logic - if `write` returns false, the data is still buffered and written. It's just a signal to the caller that any further writes should wait for a drain, which is irrelevant here since this is the last thing we're going to be writing.

Added a unit test - this test fails with the existing code (only one line is written).

Fixes #270 and #385

### DIFF
--- a/lib/appenders/dateFile.js
+++ b/lib/appenders/dateFile.js
@@ -82,13 +82,9 @@ function shutdown(cb) {
   }
 
   return openFiles.forEach((file) => {
-    if (!file.write(eol, 'utf-8')) {
-      file.once('drain', () => {
-        file.end(complete);
-      });
-    } else {
+    file.write('', 'utf-8', () => {
       file.end(complete);
-    }
+    });
   });
 }
 
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -113,13 +113,9 @@ function shutdown(cb) {
   }
 
   return openFiles.forEach((file) => {
-    if (!file.write(eol, 'utf-8')) {
-      file.once('drain', () => {
-        file.end(complete);
-      });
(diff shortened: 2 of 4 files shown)


nomiddlename:
Thanks for this - could you add a test for dateFile appender as well, please?

hansonw:
Sure thing, done :)

nomiddlename:
Thanks for this - it might be a little while before it makes it out to npm, because I need to merge in my version-2 changes. If it is something you need urgently I can patch the last release. Let me know.

hansonw:
I can wait for v2. Thanks for merging!

PR #460, opened 2017-03-24
Update README.md

Jiang-Xuan:
highlight shell command:)


nomiddlename:
Thanks for picking that up!

Issue #457, opened 2017-03-22
pattern %z (process id) can not be padded/truncated


jpstotz:
When specifying a pattern like `%10z` to insert the process id into the log file the padding to 10 characters is ignored.

I assume that the `pid` is handled internally as number and not as string, hence padding and truncation do not work. 

For example if the pattern `%10.10z` is used the following exception is thrown:

> TypeError: toTruncate.substring is not a function
>     at truncate (node_modules\log4js\lib\layouts.js:290:25)
>     at truncateAndPad (node_modules\log4js\lib\layouts.js:318:19)
>     at node_modules\log4js\lib\layouts.js:343:28
>     at Logger.<anonymous> (node_modules\log4js\lib\appenders\console.js:8:16)

My current workaround is to define my own pid as string:

```
pattern: '%6x{pid} [%r] [%[%5.5p%]] - %m',
tokens: {
	'pid': function () {
		return process.pid.toString();
	}
}
```

nomiddlename:
Release in version 2.0.0 (if not before)

PR #465, opened 2017-03-27
bugfix for issue #457

jpstotz:
bugfix for issue #457 "pattern %z (process id) can not be padded/truncated"
see https://github.com/nomiddlename/log4js-node/issues/457


nomiddlename:
Thanks for fixing this. It will be included in the 2.x release (soon).

PR #488, opened 2017-06-17
refactor: refactor setLevelForCategory code

wodog:
this one line code is enough


nomiddlename:
Thanks for this - the reason that function is more complicated is to handle setting levels on sub-categories when there is only config for the parent category. Setting a child's level should not affect the parent level. Your PR has highlighted the fact that I don't have a test that covers this situation, so thanks - I've added one now.

Issue #130, opened 2013-04-19
NDC Support


benshort:
Hi, 

Are there plans to support nested diagnostic contexts?

nomiddlename:
Something like this will be possible in version 2.x - a logger can have context added to it, which will be added to all log events that use that logger. Also log4js no longer keeps references to loggers.

jorgemsrs:
I have a patch to land MDC support for the 1.x branch. Would it be something you would be willing to accept?
Given that the 2.x branch is being developed in master how should I proceed with the PR?

nomiddlename:
You should be able to make the same changes to pattern layout in 2.x - the 2.x Logger has support for a context that works much the same as the MDC in that article. Each logEvent will get a copy of the context when it is created. The logFaces appenders make use of the values, so that should give you an example of how to access the fields. I'd rather the feature was added to 2.x - I'm planning to release it to npm very soon (just got a performance problem to look into).

Thanks for looking into this, by the way - I'm grateful for any help. If it really has to be included in 1.x for you (perhaps waiting and then upgrading is not an option), let me know and we can work something out.

jorgemsrs:
I can wait for the new release :)
Adapted the patch to fit into master. Please have a look.

Thanks

nomiddlename:
Released in version 2.0.0

PR #489, opened 2017-06-17
Add support for MDC in pattern layout

jorgemsrs:
Vaguely related to #130.


nomiddlename:
Thanks for this - very helpful.

Issue #490, opened 2017-06-17
Console appender performance improvement


jorgemsrs:
The console appender currently delegates all write operations to `console.log`. The current implementation uses `util.format` which introduces a slight performance penalty (https://github.com/nodejs/node/blob/master/lib/console.js#L82). 

I propose change the delegation to `process.stdout` which is a `WritableStream` and, thereof, is the lower level and leaner substitute.

For nodejs 8.1.1 the figures are:
benchConsoleLog*100000: 7270.032ms
benchProcessWriteIgnoreErrors*100000: 3207.920ms
benchProcessWriteCatchErrors*100000: 5114.261ms

For nodejs 6.11.0:
benchConsoleLog*100000: 4435.085ms
benchProcessWriteIgnoreErrors*100000: 3712.785ms
benchProcessWriteCatchErrors*100000: 6784.424ms

I've created a gist do showcase this behaviour. Please refer to `benchProcessWriteIgnoreErrors` for the desired and equivalent substitute:
https://gist.github.com/jorgemsrs/4a5aa24cc89f13027ea27f2d1b9c517f

nomiddlename:
Closing this - stdout appender already exists.

PR #491, opened 2017-06-17
perf(#490): Performance improvement on console appender

jorgemsrs:
Please see https://github.com/nomiddlename/log4js-node/issues/490 for
more details.

### DIFF
--- a/lib/appenders/console.js
+++ b/lib/appenders/console.js
@@ -1,10 +1,10 @@
 'use strict';
 
-const consoleLog = console.log.bind(console);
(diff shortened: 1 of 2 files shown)


nomiddlename:
Hi - there's already a `stdout` appender, created specifically to address the performance problems with using `console.log`. The `console` appender still exists mainly because people are using log4js in the browser (via browserify and things), where stdout is not available.

jorgemsrs:
I was unaware of the stdout appender. I'll close this MR as it brings no value.

Cheers

PR #505, opened 2017-07-15
Update faq.md

xrian:
LogLevelFilter,there are only four options
type、appender、level、maxLevel
minLevel does not exist.It's a mistake here
Because of this, troubled me for a long time


nomiddlename:
Thanks for this - and sorry about causing you confusion.

PR #524, opened 2017-08-17
data option to loggly appender

FDIM:
I was missing an ability to add extra fields to 'json' object visible in loggly interface as illustrated below:
image

By using configuration below I could now add branch and task fields and see them in loggly:
````
loggly: {
  type: './lib/loggly-appender', // using custom one for now
  token: process.env.LOGGLY_TOKEN,
  subdomain: process.env.LOGGLY_SUBDOMAIN,
  tags: ['web'],
  data: { branch:'develop', task:'web' },
  json: true
}
````

There shouldn't be any breaking changes, so I hope you will consider merging this PR.

Would you consider using alternative loggly client in the future? For example the one that is buffered and not dispatched right away with each event.

### DIFF
--- a/docs/loggly.md
+++ b/docs/loggly.md
@@ -8,6 +8,7 @@ Sends logging events to [Loggly](https://www.loggly.com), optionally adding tags
 * `token` - `string` - your really long input token
 * `subdomain` - `string` - your subdomain
 * `tags` - `Array<string>` (optional) - tags to include in every log message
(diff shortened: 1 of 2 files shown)


nomiddlename:
The loggly appender has moved to https://github.com/log4js-node/loggly, and it now uses a different client. If you're still interested in making this change, could you take a look at that repo, please? I'm going to close this.

PR #529, opened 2017-08-24
fix worker will die if the master exit

harlentan:


peteriman:
From the original code snippet, it is using `net.Socket`.
```javascript
    socket = net.createConnection(config.loggerPort || 5000, config.loggerHost || 'localhost');
```
> ### net.createConnection()
> A factory function, which creates a new [net.Socket](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#class-netsocket), immediately initiates connection with [socket.connect()](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketconnect), then returns the `net.Socket` that starts the connection.
>
> _(src: https://nodejs.org/dist/latest-v16.x/docs/api/net.html#netcreateconnection)_
---
Based on `net.Socket` API:
> ### Event: `'error'`
> Added in: v0.1.90
> - [\<Error\>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)
>
> Emitted when an error occurs. The `'close'` event will be called directly following this event.
>
> _(src: https://nodejs.org/dist/latest-v16.x/docs/api/net.html#class-netsocket)_
---
It seems like the behavior did not change and that the `'close'` event will still get called.

It is strange that adding the `'error'` event listener will fix the issue of workers dying when the master exit.

@harlentan, would you happen to have the replication steps for the error or any automated tests?
It will greatly help to validate this fix.

peteriman:
Similar to #1089, it's due to the way EventEmitter works.

If there is no listener to the `'error'` event, the `EventEmitter` additionally throws the `error`, which in turn becomes `uncaughtException` that terminates the Node.js process. This portion has been overlooked and this PR patches it.

> # Events
> ### Error events
> When an error occurs within an `EventEmitter` instance, the typical action is for an `'error'` event to be emitted. These are treated as special cases within Node.js.
> 
> 
> ```
> const myEmitter = new MyEmitter();
> myEmitter.emit('error', new Error('whoops!'));
> // Throws and crashes Node.js
> ```
> To guard against crashing the Node.js process the [domain](https://nodejs.org/docs/latest/api/domain.html) module can be used. (Note, however, that the `domain` module is deprecated.)
> 
> As a best practice, listeners should always be added for the `'error'` events.
> 
> ```
> const myEmitter = new MyEmitter();
> myEmitter.on('error', (err) => {
>   console.error('whoops! there was an error');
> });
> myEmitter.emit('error', new Error('whoops!'));
> // Prints: whoops! there was an error
> ```
> 

lib/appenders/multiprocess.js
```diff
@@ -128,6 +128,7 @@ function workerAppender(config) {
     socket.on('timeout', socket.end.bind(socket));
     // don't bother listening for 'error', 'close' gets called after that anyway
     socket.on('close', createSocket);
+    socket.on('error', () => {});
```
peteriman, line 131: Might want to follow #1089 and flush the buffer on `'error'` before it calls `'close'`.

```javascript
    socket.on('error', (e) => {
      debug('connection error', e);
      canWrite = false;
      emptyBuffer();
    })
```

peteriman (changes requested):
Might want to follow #1089 and flush the buffer on `'error'` before it calls `'close'`.

```javascript
    socket.on('error', (e) => {
      debug('connection error', e);
      canWrite = false;
      emptyBuffer();
    })
```

lib/appenders/multiprocess.js
peteriman, line 131: Might want to follow #1089 and flush the buffer on `'error'` before it calls `'close'`.

```javascript
    socket.on('error', (e) => {
      debug('connection error', e);
      canWrite = false;
      emptyBuffer();
    })
```

peteriman (changes requested):
Might want to follow #1089 and flush the buffer on `'error'` before it calls `'close'`.

```javascript
    socket.on('error', (e) => {
      debug('connection error', e);
      canWrite = false;
      emptyBuffer();
    })
```

Issue #447, opened 2017-03-01
logstashUDP append does not reset fields


r0bnet:
I'm not quite sure if this issue might be intended behaviour or if my understanding is lacking. I'm currently working with the log4js logstashUDP appender and i'm using it to log anything in my nodejs/express application. If an error occurrs it will not only log the error but also log some metadata like a requestId, requestBody, requestHeaders etc. This looks like this:

```js
exports.loggingErrorHandler = (err, req, res, next) => {
  logger.error('Error occurred', { requestId: req.id, headers: req.headers, body: req.body, error: err });
  next(err);
};
```
After adding those new fields to the log they will remain regardless if an error occurrs or not (only for this appender, not for e.g. console). I figured out that the logstashUDP appender does not reset those manually added fields per log. So as is said i'm not sure if this appender is intended to be used in that way. Otherwise i'm providing a solution that fits my needs (the inner log function):

```js
return function log(loggingEvent) {
    /*
     https://gist.github.com/jordansissel/2996677
     {
     'message'    => 'hello world',
     '@version'   => '1',
     '@timestamp' => '2014-04-22T23:03:14.111Z',
     'type'       => 'stdin',
     'host'       => 'hello.local'
     }
     @timestamp is the ISO8601 high-precision timestamp for the event.
     @version is the version number of this json schema
     Every other field is valid and fine.
     */

    const currentFields = Object.assign({}, config.fields);
    /* eslint no-prototype-builtins:1,no-restricted-syntax:[1, "ForInStatement"] */
    if (loggingEvent.data.length > 1) {
      const secondEvData = loggingEvent.data[1];
      for (const key in secondEvData) {
        if (secondEvData.hasOwnProperty(key)) {
          currentFields[key] = secondEvData[key];
        }
      }
    }
    currentFields.level = loggingEvent.level.levelStr;
    currentFields.category = loggingEvent.categoryName;

    const logObject = {
      '@version': '1',
      '@timestamp': (new Date(loggingEvent.startTime)).toISOString(),
      type: type,
      message: layout(loggingEvent),
      fields: currentFields
    };

    const keys = Object.keys(currentFields);
    for (let i = 0, length = keys.length; i < length; i += 1) {
      logObject[keys[i]] = currentFields[keys[i]];
    }
    sendLog(udp, config.host, config.port, logObject);
  };
```

[...]

PR #531, opened 2017-08-26
fix: sticky fields

ramicohen303:
Fixed an issue that was introduced in commit 6746e2e0488f3c900fd358af0a50d4795df65de7
All fields became sticky, so each time a field was sent via a message, it showed up in all messages after that


ramicohen303:
This is a fix for issue #447.

ramicohen303:
There is an issue with the pm2 tests, which are not related to my change, and are failing for another pull request. Looks like a couple of the tests time out.

nomiddlename:
Thanks for this, much appreciated.

Issue #532, opened 2017-08-27
LogstashUDP appender sends all arguments twice


ramicohen303:
The LogstashUDP appender takes the arguments of the log function and the config fields of the appender, and adds them to the resulting Json log object twice:
1. As child properties of the fields property.
2. As direct properties of the log object

This results in bloated messages, especially when many arguments are sent.

PR #534, opened 2017-08-27
feat(logstashUDP): Added args config option

ramicohen303:
Added a config option called args, to determine how to log args:
both (default for backward compatibility): under the fields property and as direct properties of the log object.
fields: only under the fields property.
direct: only as direct propertiesd of the log object.


ramicohen303:
A fix for issue #532

nomiddlename:
Hi, thanks for this. If you could add some tests to cover the logic around the `args` field that would be awesome, it looks quite complicated.

Thanks for your help, and sorry for taking so long to review this.

ramicohen303:
Hi

I added tests for the new args options. The failing checks are for pm2 tests, not related to my changes.
Branch is rebased.

Thank you

Rami

nomiddlename:
Awesome work, thanks a lot. Published to npm in version 2.3.9

Issue #548, opened 2017-09-21
Add flag to turn off cluster support


nomiddlename:
The new cluster support in log4js@2.x causes unexpected behaviour in environments that don't use file appenders. Some users (pm2 users in particular) were not using the file appenders, just sending output to stdout, and were happy with the output being in separate files handled by pm2 itself. Cluster support was added to remove problems with multiple processes writing to the same file at the same time. This is not a problem for these users.

I'm going to add a configuration flag to turn off the cluster support, which should revert to the previous behaviour.

nomiddlename:
See #265 for more details.

Issue #525, opened 2017-08-18
Fail on Passenger + apache : cluster mode


Benoit-Vasseur:
Hi,

We are using Log4js in our project. Works fine on our local machines.
However the code fails when it is run by apache + passenger.

We have this error : 
```
[Thu Aug 17 16:04:58.576485 2017] [core:notice] [pid 22350:tid 140174251092096] AH00094: Command line: '/usr/sbin/httpd -f /opt/application/wpt/current/apache246/conf/httpd.conf'
App 22494 stdout: 
App 22494 stderr: Trace: You required the Node Cluster module, which is incompatible with Passenger, a non-functional shim was returned and your app may still work. However, please remove the related code as soon as possible.
App 22494 stderr:     at Module.require (/usr/share/passenger/helper-scripts/node-loader.js:63:12)
App 22494 stderr:     at require (internal/module.js:20:19)
App 22494 stderr:     at Object.<anonymous> (/opt/application/Application_X/node_modules/log4js/lib/log4js.js:27:17)
App 22494 stderr:     at Module._compile (module.js:570:32)
App 22494 stderr:     at Object.Module._extensions..js (module.js:579:10)
App 22494 stderr:     at Module.load (module.js:487:32)
App 22494 stderr:     at tryModuleLoad (module.js:446:12)
App 22494 stderr:     at Function.Module._load (module.js:438:3)
App 22494 stderr:     at Module.require (module.js:497:17)
App 22494 stderr:     at Module.require (/usr/share/passenger/helper-scripts/node-loader.js:80:25)
App 22494 stderr: Trace: You required the Node Cluster module, which is incompatible with Passenger, a non-functional shim was returned and your app may still work. However, please remove the related code as soon as possible.
App 22494 stderr:     at Module.require (/usr/share/passenger/helper-scripts/node-loader.js:63:12)
App 22494 stderr:     at require (internal/module.js:20:19)
App 22494 stderr:     at Object.<anonymous> (/opt/application/Application_X/node_modules/log4js/lib/configuration.js:5:17)
App 22494 stderr:     at Module._compile (module.js:570:32)
App 22494 stderr:     at Object.Module._extensions..js (module.js:579:10)
App 22494 stderr:     at Module.load (module.js:487:32)
App 22494 stderr:     at tryModuleLoad (module.js:446:12)
App 22494 stderr:     at Function.Module._load (module.js:438:3)
App 22494 stderr:     at Module.require (module.js:497:17)
[...]

PR #550, opened 2017-09-25
fix(cluster): add flag to turn off the cluster support

nomiddlename:
Here's the change to turn off the transparent support for clustering, just in case it causes weirdness for you.


nomiddlename:
Fixes #548

nomiddlename:
Fixes #525

PR #552, opened 2017-09-27
In the above example 'tasks' does not exist

qbunt:
but 'task' does. This tripped us up while migrating an older project using this library.


nomiddlename:
Thanks for picking up that mistake.

PR #571, opened 2017-10-16
Add return types for addLayout() and shutdown() functions.

pe8ter:
Missing return types on functions causes the compiler to fail
when using the `strict` or `noImplicitAny` TypeScript flags.

### DIFF
--- a/types/log4js.d.ts
+++ b/types/log4js.d.ts
@@ -5,13 +5,13 @@ export function getLogger(category?: string): Logger;
 export function configure(filename: string): void;
 export function configure(config: Configuration): void;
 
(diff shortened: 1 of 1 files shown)


pe8ter:
I'm not certain how to interpret these test failures. There doesn't appear to be a break down of what happened on each individual test.

nomiddlename:
Yep, the travis errors are due to a flaky pm2 test, nothing to do with your changes.

nomiddlename:
Thanks for your help on this. Pushed the change to NPM in version 2.3.7

pe8ter:
Glad to help!

PR #574, opened 2017-10-25
fix: Remove setLevel from Logger TypeScript definition

hansonw:
Logger no longer has a `setLevel` method.


nomiddlename:
Thanks for this, well spotted.

PR #583, opened 2017-10-31
fix(logstashUDP): handle null and undefined in optional params

ramicohen303:


nomiddlename:
Published to npm in version 2.3.11

PR #604, opened 2017-11-09
feat(logstashUDP): Added ability to have dynamic fields

eyalmt:
Support dynamic field values, by allowing configured fields to be functions.
This code is essentially copied from the pattern layout where this functionality already exists.
It may be desirable to add it to other JSON based appenders as well.


ramicohen303:
Hi Eyal,

Can you please squash the 4 commits into one?
Can you also add a test or two for the new functionality?

Thank you

Rami

ramicohen303:
Replaced by #607

PR #605, opened 2017-11-09
chore: upgraded dev deps

ramicohen303:


nomiddlename:
Looks like the errors in node v4 and v5 are genuine - stack trace format has changed.

ramicohen303:
Will test and fix. I have not tested with older node versions.

PR #612, opened 2017-11-13
chore(lint): added linting to test files

nomiddlename:
Just noticed a few lint errors when I updated eslint.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -8,7 +8,7 @@ const debug = require('debug')('log4js:configuration');
 
 let cluster;
 try {
-  cluster = require('cluster');   // eslint-disable-line global-require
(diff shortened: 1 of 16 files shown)


nomiddlename:
Travis build is failing because it picked up the new date-format lib with the ISO8601 fix. I'll close this and fix that as well.

ramicohen303:
The changes look good to me. I assumed the test runs all lint tests as well, didn't realize they didn't cover the test folder, so good to see you fixed that too.

nomiddlename:
`npm run lint` would do both lib and tests, but `npm test` would only run eslint on the lib. I hadn't noticed the difference either. Most changes were whitespace. I'll fix the date-format stuff first, in a separate PR, then come back to this.

Issue #614, opened 2017-11-14
NSP security warning from dependencies


jornki:
# How to recreate
* Add log4js `v2.3.12` to a node project
* Run nsp against it

# Issue
* Log4js has two dependencies with DoS vulnerabilities 
   * https://nodesecurity.io/advisories/533 (timespan)
   * https://nodesecurity.io/advisories/534 (debug)

# Action
* If possible update the deps to new versions

denistn:
Hello,

could you please tell us when the new release with Pull request #616 will be published to npm? 

Thanks a lot for your work.

Denis

nomiddlename:
This should be fixed in log4js v2.4.1 - if you could confirm, that would be great.

denistn:
Hello,

based on the checks we have done, the issue is still present in log4js v2.4.1. In particular:

NSP-533
Log4js is using last version of loggly (v1.1.1), but it has an insecure dependency regarding timespan module (v2.3.0) and it is the last version. It seems timespan module is not supported anymore (last commit Aug 2016).
Is it possible to replace loggly module with another one? 


NSP-534
Log4js is using an old version of mailgun-js module (v0.7.15) instead using the last one (v0.13.1) where the security issue is resolved.

nomiddlename:
Hi, I've added NSP checking to the checks that are done in github, and NSP says there are no known vulnerabilities in the project ([NSP page for logj4s-node](https://nodesecurity.io/orgs/log4js-node/projects/2024ab95-dcc0-4cb5-9007-2d4c843988dc#open)). How are you running your report? Perhaps there are settings I need to configure on my side to reproduce the issue. Do you use the mailgun and loggly appenders in your project? They are optional dependencies and should not affect your project (or even be included) if you do not use them.

gregtillbrook:
I was just trying out snyk and it also flagged log4js as having vulnerabilities. See https://snyk.io/test/npm/log4js

nomiddlename:
Moving this to v3, where we will move the various appenders into their own projects.

jornki:
And another one (2.6.1):

* https://nodesecurity.io/advisories/664

nomiddlename:
Fixed in log4js@3.0.0

PR #616, opened 2017-11-16
fix(#614): bumped deps of debug, date-format, streamroller

nomiddlename:
I updated the deps for debug, date-format, streamroller and fixed up the tests that broke because of it. I also added the Gemfile to npmignore, and made the pm2 test potentially more reliable.


nomiddlename:
Ok, looks like I lied about making the PM2 tests more reliable :)

PR #621, opened 2017-11-25
When a file stream is full, pause output.

### LINKED ISSUE #620: Memory leak when file appender can't write
When a file appender can't write to its file, log4js leaks memory.

Steps to reproduce:

```
$ sudo touch cheese.log
$ ls -l cheese.log
-rw-r--r-- 1 root root 0 Nov 24 22:31 cheese.log
$ node test/memory-leak.js &
$ while ps -o rss,command | grep "node[ ]test/memory-leak" ; do sleep 60 ; done 
68440 node test/memory-leak.js
83408 node test/memory-leak.js
114292 node test/memory-leak.js
128812 node test/memory-leak.js
143332 node test/memory-leak.js
```

I mentioned at https://github.com/log4js-node/log4js-node/issues/575#issuecomment-346921386 that I though this might be due to `console.error` triggering a memory leak in node, but after some more testing, I see that `console.error` seems to get called just once.

By comparison, if cheese.log is writeable, there is no memory leak.

memory-leak.js:
```
"use strict";
const log4js = require('../lib/log4js');
log4js.configure({
  appenders: [{ type: 'file', filename: 'cheese.log', category: 'cheese' }]
});
const logger = log4js.getLogger('cheese');
setInterval(() => logger.info('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'), 1);
```

jsha:
WritableStream.write returns false when the stream's buffer reaches its high
water mark.
(https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback).
You can keep writing after that point, but data will just accumulate in a
buffer, producing a memory leak.

This can happen in particular when there is an error writing to a file, for
instance because the filesystem is full. Rather than leaking memory until the
process crashes, it is better to pause writes and lose some logs.

Fixes #620.

### DIFF
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -36,6 +36,7 @@ function fileAppender (file, layout, logSize, numBackups, compress, timezoneOffs
   //there has to be at least one backup if logSize has been specified
   numBackups = numBackups === 0 ? 1 : numBackups;
 
+  var paused = false;
   function openTheStream(file, fileSize, numFiles) {
     var stream;
     if (fileSize) {
@@ -56,6 +57,9 @@ function fileAppender (file, layout, logSize, numBackups, compress, timezoneOffs
     stream.on("error", function (err) {
       console.error("log4js.fileAppender - Writing to file %s, error happened ", file, err);
     });
+    stream.on("drain", function () {
+      that.paused = false;
+    });
     return stream;
   }
 
@@ -65,7 +69,12 @@ function fileAppender (file, layout, logSize, numBackups, compress, timezoneOffs
   openFiles.push(logFile);
 
   return function(loggingEvent) {
-    logFile.write(layout(loggingEvent, timezoneOffset) + eol, "utf8");
+    if (paused) {
+      return;
+    } else if (!logFile.write(layout(loggingEvent, timezoneOffset) + eol, "utf8")) {
+      process.stdout.write('write buffer full (possibly due to errors), dropping logs until it recovers.\n');
+      paused = true;
+    }
   };
 
 }


jsha:
Ah, it looks I incorrectly pulled master before starting on this, and had an old version. It looks like the relevant code is now in the `streamroller` library, but you can see the gist from the above.

lib/appenders/file.js
```diff
@@ -65,7 +69,12 @@ function fileAppender (file, layout, logSize, numBackups, compress, timezoneOffs
   openFiles.push(logFile);
 
   return function(loggingEvent) {
-    logFile.write(layout(loggingEvent, timezoneOffset) + eol, "utf8");
+    if (paused) {
```
lusarz: Wouldn't it be more elegant to write:
 
    if (!paused && !logFile.write(layout(loggingEvent, timezoneOffset) + eol, "utf8")) {
      process.stdout.write('write buffer full (possibly due to errors), dropping logs until it recovers.\n');
      paused = true;
    }
?

nomiddlename:
Removing this from the 3.0.0 milestone, could take a while to come up with the correct way to handle this issue.

jsha:
Any further thoughts on this? I'm no longer quite convinced that pausing output is the right call. If the logs are important, it's probably better to crash right away rather than continue to run in a condition where things can't be logged. But in either case, leaking memory and crashing some time later is definitely wrong.

nomiddlename:
This PR will be replaced by the work done in #938 - emitting a "log4js:pause" event and letting the caller decide what they should do.

PR #631, opened 2017-12-19
Use the "files" property to only include the `lib` folder when `npm install`ing

brettneese:
Currently, `log4js-node` installs the entire project when being `npm install`ed. This is an anti-pattern, and it leads to a package almost 1MB larger than it needs to be, as it includes the entire set of docs every time it gets installed (among other things.)

This PR reduces the size of the `npm install`ed packaged to ~100KB by only including the bits that are `log4js`. This makes the project easier to used in constrained environments.

More details: https://docs.npmjs.com/files/package.json#files

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,11 @@
 {
   "name": "log4js",
-  "version": "2.4.1",
+  "version": "2.4.2",
   "description": "Port of Log4js to work with node.",
   "homepage": "https://log4js-node.github.io/log4js-node/",
+  "files": [
+    "lib"
+  ],
   "keywords": [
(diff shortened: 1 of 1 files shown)


brettneese:
(Also, apologies for the mess. Accidentally submitted the PR before I was finished writing it!)

nomiddlename:
Thanks for this. I'll probably roll it into the next npm release.

ramicohen303:
FYI - I am including this fix in the v3 work.

nomiddlename:

brettneese:
👍

PR #636, opened 2018-01-02
chore(loggly): switch to mainted, offical loggly fork

graingert:
Fixes #635

PR #641, opened 2018-01-09
Update nodemailer

benitogf:
hello 👋 I'm getting these warnings, just checking if this will break anything here
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "url": "http://github.com/log4js-node/log4js-node/issues"
   },
   "engines": {
-    "node": ">=4.0"
+    "node": ">=6.0"
   },
   "scripts": {
     "clean": "find test -type f ! -name '*.json' ! -name '*.js' ! -name '.eslintrc' -delete && rm *.log",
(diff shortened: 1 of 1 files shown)


benitogf:
saw that the issue with node-uuid is from here https://github.com/winstonjs/node-loggly/pull/74 😃 will send a pr to update when it's merged/bumped, thanks

kristoferbaxter:
I went to make this same change today, but noticed the following in the FAQ.

https://github.com/log4js-node/log4js-node/blob/master/docs/faq.md#npm-complains-about-nodemailer-being-deprecated-what-should-i-do

> Nodemailer version 4.0.1 (the not-deprecated version) requires a node version >= 6, but log4js supports node versions >= 4. So until I stop supporting node versions less than 6 I can't update the dependency. It's only an optional dependency anyway, so you're free to install nodemailer@4.0.1 if you want - as far as I know it should work, the API looks the same to me. If you know that the smtp appender definitely doesn't work with nodemailer v4, then please create an issue with some details about the problem.

Might I suggest also changing the engine field in `package.json` so this request can be merged once the author stops supporting the LTS release of Node 4?

benitogf:
@kristoferbaxter thanks for the good information, updated the engine field as suggested

benitogf:
also think that https://github.com/log4js-node/log4js-node/pull/636 will fix the uuid warning

nomiddlename:
We will be dropping support for node v4 in version 3.x, so we will update this then.

PR #648, opened 2018-01-15
fix: deferred logger configuration.

tmihalcin:
logger will be configured either by user or at first call to getLogger.

According to documentation `If you do not call configure, log4js will use LOG4JS_CONFIG (if defined) or the default config.`
That is currently not the case as configure is called at module load time no matter if/when client code calls the configure.
This PR aims to fix this behavior.

Reasoning: we are using webpack for server packing and due to dynamic loading of appenders, these are not included in the final package (the suggestions in several bug reports don't work either).
In our project I would like to override the loading behavior to address the issue, but call to configure with the default config already tries to load the stdout which is not in the bundle and fails.

This PR preserves the original behavior without breaking backwards compatibility, but allows the client code to perform the configuration only once.

### DIFF
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -163,6 +163,10 @@ function isMaster() {
  * @return {Logger} instance of logger for the category
  */
 function getLogger(category) {
+  if (!enabled) {
+    configure(process.env.LOG4JS_CONFIG || defaultConfig);
+  }
+
   const cat = category || 'default';
   debug(`creating logger as ${isMaster() ? 'master' : 'worker'}`);
   return new Logger((isMaster() ? sendLogEventToAppender : workerDispatch), cat);
@@ -285,5 +289,3 @@ const log4js = {
 };
 
 module.exports = log4js;
-// set ourselves up
-configure(process.env.LOG4JS_CONFIG || defaultConfig);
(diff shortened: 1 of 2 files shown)


nomiddlename:
Does this change make something like this work with webpack?
```javascript
log4js.configure({
  appenders: { 'out': { type: 'console' } },
  categories: { default: { appenders: ['out'], level: 'debug' }
});
```
Will the console appender now get bundled up correctly? Or would you still need some other workarounds?

tmihalcin:
It doesn't, the problem is with dynamic appender loading. 
What would probably help is separating the appenders to separate modules.
In the meantime, this PR should align the code with documentation while maintaining bakcwards compatibility and it will help me make the webpack setup work with one additional workaround which isn't possible right now.

nomiddlename:
Cool - thanks for this. If you could add some docs on the webpack workaround to the FAQ when you get it working, that would be a big help for a lot of users.

tmihalcin:
Sure, no problem.
I see you made a new release with this fix. Could you please publish it to the npm?
Thanks a lot.

nomiddlename:
Yep, travis-ci was having a bit of a meltdown yesterday so I was waiting for the builds to run to make sure everything still works before pushing to NPM.

PR #650, opened 2018-01-16
add rabbitmq appender to push to rabbitmq

Anteoy:
Added ability to push message to rabbitmq

### DIFF
--- a/docs/appenders.md
+++ b/docs/appenders.md
@@ -40,6 +40,7 @@ The following appenders are included with log4js. Some require extra dependencie
 * [smtp](smtp.md)
 * [stderr](stderr.md)
 * [stdout](stdout.md)
+* [rabbitmq](rabbitmq.md)
 
 ## Other Appenders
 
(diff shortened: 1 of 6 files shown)


nomiddlename:
Thanks for submitting this, and for putting the effort in to write tests - much appreciated. It looks like this appender creates a new connection to rabbitmq with every log message, is there much overhead for doing that? What is the performance like?

Also, if you could add a page to the docs describing the options for the appender, that would be awesome.

Anteoy:
I am honored to submit a PR for this, I am a novice using nodejs, I spent a lot of time to adjust the code. I am using amqplib rabbitmq official recommended npm package, I wrote in accordance with the official sample program, thank you for your reminder, for each message to establish a rabbitmq tcp connection is a very bad practice. I have now updated and submitted the code. In addition, I found that after rewriting the news for the test is relatively simple, is this enough? Or what kind of a way to fake a message?

Sure, I will write the doc later

Anteoy:
I have added the doc

Anteoy:
I have solved the problem of testing messages, updated and submitted the relevant code.

nomiddlename:
Awesome, thanks for your hard work.

nomiddlename:

PR #663, opened 2018-02-12
Configure method for oneliners

idenisovs:
The changes within this PR allow to write code like:
```javascript
const log = require('log4js').configure(conf).getLogger('daemon');
```

For this PR I have:

1. Added `return` statement for log4js.configure(...) method.
2. Refactored the `configuration-test.js` file.
3. Added new test to `configuration-test.js` file. This test checks  the value, returned by `log4js.configure(...)` method.
4. Appended the `configure(...)` related article in the docs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,6 +10,8 @@ If you are using `cluster`, then include the call to `configure` in the worker p
 
 Configuration objects must define at least one appender, and a default category. Log4js will throw an exception if the configuration is invalid.
 
+`configure` method call returns the configured log4js object.
+
 ### Configuration Object
 Properties:
 * `levels` (optional, object) - used for defining custom log levels, or redefining existing ones; this is a map with the level name as the key (string, case insensitive), and an object as the value. The object should have two properties: the level value (integer) as the value, and the colour. Log levels are used to assign importance to log messages, with the integer value being used to sort them. If you do not specify anything in your configuration, the default values are used (ALL < TRACE < DEBUG < INFO < WARN < ERROR < FATAL < MARK < OFF - note that OFF is intended to be used to turn off logging, not as a level for actual logging, i.e. you would never call `logger.off('some log message')`). Levels defined here are used in addition to the default levels, with the integer value being used to determine their relation to the default levels. If you define a level with the same name as a default level, then the integer value in the config takes precedence. Level names must begin with a letter, and can only contain letters, numbers and underscores.
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -230,6 +230,8 @@ function configure(configurationFileOrObject) {
   }
 
   enabled = true;
+
+  return log4js;
 }
 
 /**
--- a/test/tap/configuration-test.js
+++ b/test/tap/configuration-test.js
@@ -3,54 +3,79 @@
 const test = require('tap').test;
 const sandbox = require('sandboxed-module');
 
-test('log4js configure', (batch) => {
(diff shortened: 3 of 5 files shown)


nomiddlename:
Thanks for this - only one small thing left to do, which is update the typescript definition (it's in `types/log4js.d.ts`) to include the return value.

idenisovs:
@nomiddlename Sorry for asking this (I have no experience with TypeScript), but will be type `object` the proper solution for this?

Like 

```typescript
export function configure(filename: string): object;
export function configure(config: Configuration): object;
```

nomiddlename:
I have no idea either - all I know is that the TypeScript people are pretty quick off the mark if we get it wrong :) - but I think you want to use `this` instead of `object` in your example above, to show it is returning a log4js class (sort of). A good way to test it is to add your one-liner code to the test.ts file in the types directory and see if it compiles with typescript.

idenisovs:
Usage of `this` produces the following errors:

```
$ tsc test.ts
log4js.d.ts(5,46): error TS2526: A 'this' type is available only in a non-static member of a class or interface.
log4js.d.ts(6,51): error TS2526: A 'this' type is available only in a non-static member of a class or interface.
```
Seems, that with current implementation of `test.ts` file, there is only two solutions:

1. Redefine internal interface `Log4js`, which will have copy of exported functions itself (so it will be declared here twice).
2. Set return type of `configure(...)` methods to `any`.

At the moment, first seems the better, so I'll put it in the definition file, and will hope, that later someone, more experienced with Typescript, will put here the better solution.

nomiddlename:
Hey typescript experts (or at least people who have contributed changes to the typescript definitions): @xialeistudio @pe8ter @hansonw @taoqf @AmShaegar13  - could you offer some guidance to @idenisovs please? I'm out of my depth here.

AmShaegar13:
How about this:

```
diff --git types/log4js.d.ts types/log4js.d.ts
index 01c48eb..f6822ec 100644
--- types/log4js.d.ts
+++ types/log4js.d.ts
@@ -2,8 +2,8 @@
 
 export function getLogger(category?: string): Logger;
 
-export function configure(filename: string): void;
-export function configure(config: Configuration): void;
+export function configure(filename: string): Log4js;
+export function configure(config: Configuration): Log4js;
 
 export function addLayout(name: string, config: (a: any) => (logEvent: LoggingEvent) => string): void;
 
@@ -13,6 +13,15 @@ export function levels(): Levels;
 
 export function shutdown(cb?: (error: Error) => void): void | null;
 
+export interface Log4js {
+	getLogger;
+	configure;
+	shutdown;
+	connectLogger;
+	levels;
+	addLayout;
+}
+
 export interface BaseLayout {
 	type: 'basic';
 }
diff --git types/test.ts types/test.ts
index 91e1b84..e8159a6 100644
--- types/test.ts
+++ types/test.ts
@@ -1,6 +1,6 @@
 import * as log4js from './log4js';
 
-log4js.configure('./filename');
+log4js.configure('./filename').getLogger();
 const logger1 = log4js.getLogger();
 logger1.level = 'debug';
 logger1.debug("Some debug messages");
```

idenisovs:
@AmShaegar13 I spent couple of hours, trying to figure out, how to do something like this. Nevermind, that it could be so easy. 😆 
Thank you!

nomiddlename:
Thanks for this @idenisovs, and thanks to @AmShaegar13 for helping out. This will go out to NPM in version 2.6.x, I'd like to bundle up deprecating a few of the appenders first in preparation for removing them in version 3.x.

PR #672, opened 2018-02-27
Added Slack Webhook appender

Thor-:


nomiddlename:
Thanks for this. I have a couple of issues before I can accept:

- needs tests, I need to make sure that future changes won't break the appenders and I can't test them all manually
- needs documentation, explaining how to use it and when you should use it instead of the other slack appender we have
- needs to conform to the eslint rules

Thanks for wanting to contribute, let me know if you need more explanation.

nomiddlename:
I'm going to close this one. If you want to revisit it, then raise the PR against the [slack appender repo](https://github.com/log4js-node/slack)

PR #677, opened 2018-03-04
fixes #675 - added docs blurb about delegation to util.format

### LINKED ISSUE #675: Docs don't specify that util.format is used to format messages
Having never used a node logger and trying to use this one, I could not find anywhere in the documentation anywhere or in any examples that explained what formatting placeholders were available (`%s`, `%d`, etc)

I basically gave up on this library simply because of that, giving me a bad impression for what I assumed to be a core feature.  While reading the winston docs they made explicit mention that formatting was done via `util.format`, so I came back here and found in the source that it seems the same is true in log4js as well

The docs should really make mention of this, without prior knowledge there's no point of reference that this would be the case

alex-sherwin:
Added a blurb to the API about delegation to the node [`util.format`](https://nodejs.org/api/util.html#util_util_format_format_args) for formatting log string placeholders like `%s`, `%d` etc.

Without any prior knowledge that this might be the case, the docs currently don't make any mention of it


nomiddlename:
Thanks for the PR - I think what you've said is true, but it is layout-dependent. The basic, coloured, and message pass-through layouts will do this, but not the pattern layout.

PR #690, opened 2018-03-24
Version 3.x

lykesun:
rabbit appender add vistual host configuration


nomiddlename:
Hi, this PR looks like you want to merge the version-3.x branch into master - which I'm guessing is not what you intended. If you've got a change that's just for the rabbitMQ appender, then it's probably best to make that change from master (in your forked repo, branch from master, not version-3.x) and then re-submit your PR. Remember to include tests that cover your change, and update any documentation that needs it. Thanks again for wanting to help out.

Issue #570, opened 2017-10-16
Automated test for typescript definition


nomiddlename:
We need a way of making sure we don't break the typescript definitions again. I don't use typescript, so I'd be very happy to get some input from people who do. @taoqf @AmShaegar13 @xialeistudio 

Is just running the `types/test.ts` file and not getting any errors enough? I should be able to hook that into the travis build pretty easily.

taoqf:
I think it will be enough if all tests are included in types/test.ts.

xialeistudio:
yes,I commit a copy to [DefinitelyTyped/DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped),
It passed the test

pe8ter:
I submitted #710 which automatically verifies that the typings are strict on prepush. What this does not do is validate that the typings adhere to the library's actual functionality which cannot trivially be done without actually writing the library itself in TypeScript, so I don't know if the PR addresses the concern of this issue.

nomiddlename:
Closing this one. The typescript test we have now is good enough.

Issue #709, opened 2018-05-12
log4js.d.ts does not follow `strict` TypeScript typings


pe8ter:
### Overview

The typings in version `2.6.0` do not build successfully when using the `strict` TypeScript compiler option.

### Steps to recreate

1. Use this _tsconfig.json_:

```
{
  "compilerOptions": {
    "strict": true
  }
}
```

2. Use this _main.ts_:

```
import * as log4js from 'log4js';
```

3. Compile:

```
$ tsc
```

### Expected result

The compiler should emit no errors.

### Observered result

The compiler emits these errors:

```
node_modules/log4js/types/log4js.d.ts(4,2): error TS7008: Member 'getLogger' implicitly has an 'any' type.
node_modules/log4js/types/log4js.d.ts(5,2): error TS7008: Member 'configure' implicitly has an 'any' type.
node_modules/log4js/types/log4js.d.ts(6,2): error TS7008: Member 'addLayout' implicitly has an 'any' type.
node_modules/log4js/types/log4js.d.ts(7,2): error TS7008: Member 'connectLogger' implicitly has an 'any' type.
node_modules/log4js/types/log4js.d.ts(8,2): error TS7008: Member 'levels' implicitly has an 'any' type.
node_modules/log4js/types/log4js.d.ts(9,2): error TS7008: Member 'shutdown' implicitly has an 'any' type.
```

pe8ter:
I can put together a PR to do this. Would you be opposed to me adding `typescript` as a dev dependency so that these typings can be checked via an npm script?

PR #710, opened 2018-05-12
fix(#709): strict Log4js typings

pe8ter:
* Strictly type the exported Log4js interface.
* Test the typings for strictness using the TypesScript compiler.

The PR addresses #709.

A few things to note:

* I did not see a preferred version of npm to use when installing new dependencies. I use version `5.6.0` which makes some significant changes to the checked-in _package-lock.json_. If you'd prefer a different version of npm to tamp down those changes, I'd be happy to use it and squash the history.
* I added TypeScript as a dev dependency and an npm script to test the typings, but I did not prescribe how that script runs automatically. If you'd like to see that in prepush or elsewhere I can make that change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8570,6 +8570,12 @@
       "integrity": "sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=",
       "dev": true
     },
+    "typescript": {
+      "version": "2.8.3",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-2.8.3.tgz",
+      "integrity": "sha512-K7g15Bb6Ra4lKf7Iq2l/I5/En+hLIHmxWZGq3D4DIRNFxMNV6j2SHSvDOqs2tGd4UvD/fJvrwopzQXjLrT7Itw==",
+      "dev": true
+    },
     "uglify-js": {
       "version": "2.8.29",
       "resolved": "https://registry.npmjs.org/uglify-js/-/uglify-js-2.8.29.tgz",
--- a/package.json
+++ b/package.json
@@ -29,11 +29,12 @@
   },
   "scripts": {
     "clean": "find test -type f ! -name '*.json' ! -name '*.js' ! -name '.eslintrc' -delete && rm *.log",
-    "prepush": "npm test",
+    "prepush": "npm test && npm run typings",
     "commitmsg": "validate-commit-msg",
     "posttest": "npm run clean",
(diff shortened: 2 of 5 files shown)


nomiddlename:
This is great - thanks for doing this, this fixes #570. I don't use typescript, so it's good to have this automated.

Package-lock.json was last generated with npm v6.0.1, which is probably the reason for the large number of changes.

I'd definitely want this to run automatically - so prepush is probably a good choice.

pe8ter:
Okay. Changes made, history squashed, and ready for review.

theanurin:
@nomiddlename Hi, I'm a user of the log4js. When can I expect this fix will be published into NPM? We switching our project to strict build. The issue is blocked our activity. Also we do not want to create a copy of the package with this fix in NPM....
Thanks a lot

nomiddlename:
Published to npm in log4js@2.6.1

nicojs:
Awesome work! Thanks everyone!

pe8ter:
I just stumbled across these Microsoft TypeScript helper modules:

* [dtslint](https://github.com/Microsoft/dtslint): Lint existing typings. This could replace the TypeScript compiler step added in the fix for this issue with a little bit more "official" way to verify typings.
* [dts-gen](https://github.com/Microsoft/dts-gen): Generate typings from JavaScript. It's not bulletproof, but it may offer some insight into how to type `log4js` in the future.

I'll take a look when I have some free time and see if either of these could help in any way.

nomiddlename:
Thanks @pe8ter - that would be a great help.

pe8ter:
`dts-gen` ended up not being much help. The package advertises itself as a starting point for creating typings from plain JavaScript and the Log4js typings are already much more advanced than what the tool generates.

See [this commit](https://github.com/pe8ter/log4js-node/commit/d93a5c3c33a34cc4f2662a3f8458e1a4aeadddfb) for info about `dtslint`. There's a large comment at the bottom.

nomiddlename:
That looks pretty handy - anything that gives suggestions will be useful for people who don't use typescript but want to support it (like me). If you want to switch log4js over to using that it's ok by me.

pe8ter:
I'm experiencing [this bug](https://github.com/Microsoft/dtslint/issues/110) with the `dtslint` library. I'm going to wait a bit to see if there's any movement on the issue, and if not, submit a PR to that project with my fix. I know they will have an issue with my solution since it can't be tested (as far as I know) with their current test rig.

PR #714, opened 2018-05-15
fix(#713): handle unexpected hangups in multiprocess appender

### LINKED ISSUE #713: [multiprocess appender]: Master crashes when a worker process crashes
The host process of the log4js appender `'multiprocess'` appender with `role: 'master'` is in risk of crashing. 

If a connected logger (`role: 'worker'`) exits (or crashes) **without** calling `log4js.shutdown()`, it results in an [uncaught exception](https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly) on the host process. If that is not handled, the host process exists:

```
      ^

Error: read ECONNRESET
    at TCP.onread (net.js:659:25)
Emitted 'error' event at:
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:59:3)
    at process._tickCallback (internal/process/next_tick.js:174:19)
```

My suggestion is to add a listener: `clientSocket.on('error', handleError);`. The handleError method can be implemented with a `debug()` log message IMO. Don't know what else to do. If you agree, I'm prepared to do the work and create a PR.

Although I agree worker processes should call `shutdown` like good students, the penalty shouldn't be a crashing process. Expecting applications to implement the `process.on('uncaughtException')` is not a valid solution in my opinion, as stated in the documentation:

> The correct use of 'uncaughtException' is to perform synchronous cleanup of allocated resources (e.g. file descriptors, handles, etc) before shutting down the process. It is not safe to resume normal operation after 'uncaughtException'.

Tested on node 10.0.0.

nicojs:
Handle the `'error'` event on client sockets of worker logger.

Fixes #713

### DIFF
--- a/lib/appenders/multiprocess.js
+++ b/lib/appenders/multiprocess.js
@@ -64,8 +64,21 @@ function logServer(config, actualAppender, levels) {
       }
     }
 
+    function handleError(error) {
+      const loggingEvent = {
+        startTime: new Date(),
+        categoryName: 'log4js',
+        level: levels.ERROR,
+        data: ['A worker log process hung up unexpectedly', error],
+        remoteAddress: clientSocket.remoteAddress,
+        remotePort: clientSocket.remotePort
+      };
+      actualAppender(loggingEvent);
+    }
+
     clientSocket.on('data', chunkReceived);
     clientSocket.on('end', chunkReceived);
+    clientSocket.on('error', handleError);
   });
 
   server.listen(config.loggerPort || 5000, config.loggerHost || 'localhost', function () {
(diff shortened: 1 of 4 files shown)


nicojs:
Fixed the code analysis warnings and did a rebase.

nicojs:
Hmm the test I've written seems to only work on windows.

@nomiddlename it seems that my way of forcing an error event on the `'master'` (killing an active worker) only works on windows. I've recreated the travis build locally on linux and I can confirm the 'error' event is never raised.

What do you want me to do? I can rewrite the tests more like a unit test, raising the 'error' event directly. Or I can remove the specific assertions on the exact messages.

nicojs:
I ended up doing both, created 2 tests. Tests are green 🍏 📗 💚

nomiddlename:
Awesome work, thanks!

nomiddlename:
Published to NPM in log4js@2.7.0

PR #717, opened 2018-05-28
fix(types): loose type restrict in Logger

Brooooooklyn:


nomiddlename:
Published to NPM in version 2.8.0

PR #724, opened 2018-06-02
Making Logger def a class instead of interface.

vangorra:
Doing this allows for typescript and angular to use the class as a dependency injection object.


nomiddlename:
Thanks for this.

nomiddlename:
published to npm in version 2.9.0

PR #762, opened 2018-07-20
Fix duplicate log messages in TCP server (fixes #761)

### LINKED ISSUE #761: TCP server produces duplicate log messages
`dataSoFar` in [tcp-server.js](https://github.com/log4js-node/log4js-node/blob/v3.0.1/lib/appenders/tcp-server.js) is not reset to an empty string when events are processed.

I'll create a PR to fix this issue.

florianreinhart:
This also fixes a potential concurrency issue by storing received data per socket connection instead of globally.


florianreinhart:
Fixes #761.

nomiddlename:
Thanks for fixing my dodgy code!

PR #763, opened 2018-07-24
Document log-format function, allow it in TypeScript

philipmw:


nomiddlename:
Thanks for doing this - could you add a line or two to the `test.ts` file that covers this, please? It just tries to compile the ts file, but it would help to make sure we don't break this again later on.

philipmw:
@nomiddlename , done, thanks for the suggestion.

PR #771, opened 2018-08-02
Change for good cheese

bertho-zero:
Comté > Gouda


nomiddlename:
Thanks for this important improvement.

Issue #772, opened 2018-08-08
pattern layout function tokens broken in 3.0.4


nomiddlename:
The JSON serialisation/de-serialisation breaks the user-defined token functions in pattern layout. (Raised by DP on the Slack channel).

nomiddlename:
Published a fix to npm in version 3.0.5. Please re-open if the issue persists.

PR #773, opened 2018-08-08
fix(#772): removed serialisation, replaced with freezing

nomiddlename:
Rather than use JSON serialising and deserialising to copy an immutable config object (if it comes from node-config), which causes problems for the pattern layout user-defined functions, I've made it so that the config passed to appenders is always immutable and they will have to deal with that. This only caused problems for the dateFile and multiFile appenders in the core. Optional appenders may have problems with it (GELF, logstashUDP, and smtp appenders will all need fixing). 

Because this is potentially a breaking change, I'll bump up the version number to 4.x. I won't push this out until I've updated the GELF, logstashUDP and SMTP appenders to work with it.

### DIFF
--- a/docs/writing-appenders.md
+++ b/docs/writing-appenders.md
@@ -17,7 +17,7 @@ If that fails, an error will be raised.
 ## Appender Modules
 
 An appender module should export a single function called `configure`. The function should accept the following arguments:
-* `config` - `object` - the appender's configuration object
+* `config` - `object` - the appender's configuration object, note that this will be immutable (Object.freeze)
 * `layouts` - `module` - gives access to the [layouts](layouts.md) module, which most appenders will need
   * `layout` - `function(type, config)` - this is the main function that appenders will use to find a layout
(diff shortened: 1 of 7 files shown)


nomiddlename:
Looks like my cunning use of the object assignment rest parameter thing is not supported in node 6.

victor0801x:
I'm sorry, I cause this issue.
can we use deep-clone lib for quick fix first ?

```js
const deepFreeze = require("deep-freeze");
const clone = require("rfdc")({ proto: true, circles: true });

const ori = {
	hi: () => console.log("hi"),
	arr: ["a"],
	obj: { key: "value" },
	alwaysIncludePattern: false,
};
const f = deepFreeze(ori);
f.hi();
//f.hi = () => console.log("hello");
//f.alwaysIncludePattern = true;
//f.arr.push("b");
//f.obj.key2 = "value2";

const c = clone(f);
c.hi(); c.hi = () => console.log("hello");
c.hi();
c.alwaysIncludePattern = true
c.arr = 1;
console.log(c)
```

nomiddlename:
I'll close this PR, it causes more problems than it fixes.

Issue #687, opened 2018-03-21
log4js.shutdown typedef is wrong?


kurogelee:
log4js.shutdown typedef seems to be as follows.
`export function shutdown(cb: (error?: Error) => void): void | null;`

* current types/log4js.d.ts
`export function shutdown(cb?: (error: Error) => void): void | null;`

When I called "log4js.shutdown();" in TypeScript, following error was displayed.

```
TypeError: cb is not a function
    at WriteStream.complete (C:\research\2018\ts-sample\node_modules\log4js\lib\log4js.js:262:7)
```

nomiddlename:
Fixed in log4js@3.0.6

PR #779, opened 2018-08-13
fix(#687): shutdown callback arg is not optional

nomiddlename:


nomiddlename:

PR #780, opened 2018-08-17
fix require.main for use with Node.js mjs modules

t4d-io-website:
When using the experimental ES2015 support with Node.js (mjs modules) the main property of require is undefined causing the third check to fail to loading appenders to always fail. This simple fix will enable the fourth check to execute when loading mjs modules.


nomiddlename:
Thanks for this, much appreciated.

nomiddlename:

PR #790, opened 2018-09-25
refactor(multiprocess): serialisation consistency

LoneRifle:
Delegate all serialisation tasks to LoggingEvent, instead of
maintaining a separate codepath in the multiprocess appender


nomiddlename:
Thanks for this - I really should remove/deprecate this appender now that the tcp appender exists, but your PR is a good change to make in the interim.

Issue #793, opened 2018-09-28
multiFile appender's call back for shutdown() never called if it has no underlying file appender


so5:
Hi, I found a possibly bug in appenders/multiFile.js. 

multiFile appender's shutdown() only calls underlying file appenders' shutdwon().
So, if no file appender is used before calling shutdown(), the callback routine never called.

I already fix this problem and I will make a PR shortly after.

** code for reproduce **

```
const log4js = require("log4js");

log4js.configure({
  appenders: {
    multi: { type: 'multiFile', base: 'logs/', property: 'categoryName', extension: '.log' }
  },
  categories: {
    default: { appenders: [ 'multi' ], level: 'debug' }
  }
});
log4js.shutdown((err)=>{
  if(err) console.log("ERROR:", err);
  console.log("this line should be called");
});
```

** expected result **
```this line should be called``` should be printed on console

** actual result ***
no output

PR #794, opened 2018-09-28
fix: #793

so5:


nomiddlename:
Thanks for this, good catch - could you add a test that covers the change, please? That way I can be sure it won't be broken again by any future changes.

so5:
Hi, I add a test case which will call shutdown before use.

nomiddlename:
Thanks again, I appreciate the help.

nomiddlename:

PR #804, opened 2018-11-04
fix(#789): use flatted in place of CircularJSON

### LINKED ISSUE #789: Replace circular-json with flatted
CircularJSON has been marked deprecated in npm, in favour of a new library from the same author: flatted. 

flatted uses a different serialization schema from CircularJSON, and as a result, moving to flatted may be a breaking change for those who use the recording appender and are trying to replay older files. 

This may be mitigated if a change to flatted allows for straight-up serialization/deserialization of non-circular structures.

This issue hence proposes the replacement of CircularJSON with flatted, corresponding adjustments to tests, and a new major version release.

LoneRifle:
CircularJSON is now deprecated, so replace it with its successor
and use it in the sole location where we used CircularJSON, ie
in the LoggingEvent serialization/deserialization functions

* npm - install flatted, remove circular-json
* LoggingEvent - use flatted for serialization/deserialization
* tests - acknowledge that flatted (and circular-json) should not
  parse anything other than their respective serialized payloads
  and use flatted when parsing/stringifying test data

Fixes #789

### DIFF
--- a/lib/LoggingEvent.js
+++ b/lib/LoggingEvent.js
@@ -1,4 +1,4 @@
-const CircularJSON = require('circular-json');
+const flatted = require('flatted');
 const levels = require('./levels');
 
 /**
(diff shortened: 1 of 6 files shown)


nomiddlename:
Thanks for this. As this is a (potentially) breaking change, I'm going to schedule it for 4.0.0. I'll bundle in the changes to the streamroller library as well, since they're pretty major.

Issue #640, opened 2018-01-08
FeatureRequest: nolog in general config


avivacaller:
It would be great if we could define nolog regex in `log4js.configure()` and not just when using connect with express. Is this possible?

nomiddlename:
It's not possible at the moment, but you could implement this in a similar way to the `categoryFilter` and `logLevelFilter` appenders. They wrap around other appenders, and filter what gets sent to them. The new filter would filter the log event data instead of just checking the category or log level.

mattiacapitanio:
Hi, I downloaded the code (commit f7eee87) and the feature is still not present. 
If no one is following it, I would like to work on this.

nomiddlename:
Go for it @mattiacapitanio - just follow these [guidelines for contributing](https://log4js-node.github.io/log4js-node/contrib-guidelines.html) and you'll be fine.

mattiacapitanio:
Thank you @nomiddlename for your indications and suggestions. 
I just submitted my changes for review.

Issue #684, opened 2018-03-09
the param  daysToKeep  don't work


bobozuyuan:
the param  daysToKeep  don't work,  then files older than that many days didn't be deleted

cula:
Please offer the `fs.statSync` (just using node in the dir) results of the files which should have been deleted.

cula:
I've recoded this part of code and will be released soon

PR #815, opened 2018-12-02
FeatureRequest: nolog in general config #640

mattiacapitanio:
Added the feature to exclude the log events depending on the regular expression provided in the configuration.
The feature works in a similar way to the categoryFilter and logLevelFilter appenders. The log events, which will match the regular expression, will be excluded and not passed to the appender.
The regular expressions can be configured through a string or an array of strings. 
The log events will be evaluated in case-insensitive mode.

### DIFF
--- a/docs/noLogFilter.md
+++ b/docs/noLogFilter.md
@@ -1,0 +1,58 @@
+# Category Filter
+
+The no log filter allows you to exclude the log events that an appender will record. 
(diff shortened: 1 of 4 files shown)


nomiddlename:
This looks good - thanks for the high quality contribution. I'll merge it into master, but it won't get released to NPM until I've finished the changes needed for #684 (amongst others) in version 4.0.0

mattiacapitanio:
Great! Thank you

PR #819, opened 2018-12-19
Feature - Add status code rulesets for connectLogger

blakambr:
When using the `connectLogger` and the level set to `auto`, logging levels are set accordingly:
* http responses 3xx, level = `WARN`
* http responses 4xx & 5xx, level = `ERROR`
* else, level = `INFO`

This general grouping does not always work however as codes such as `304 Not Modified`, which in some cases are expected, are always returned as `WARN` and cannot be assigned to `INFO` without changing it for all other codes.

This feature aims to bring in some customisability around how those levels are set by passing in `statusRules` via the `options` object. An example would be:

```javascript
app.use(log4js.connectLogger(logger, { level: 'auto', statusRules: [
  { from: 200, to: 299, level: 'debug' }, // all 2xx are logged at DEBUG
  { codes: [304],  level: 'info' }        // 304 is logged at INFO
]}));
```

There are two types of rulesets that can be defined:
  1. **Range:** All codes between `from` and `to` are set to level set on the rule
  2. **Contains:** Any specified codes are set to level set on the rule

If a status code does not match any user-defined rule, it will simply use the default level outlined either by `auto` or user-specified level.

Additional tests have been added to `connect-logger-test.js`

### DIFF
--- a/docs/connect-logger.md
+++ b/docs/connect-logger.md
@@ -30,6 +30,7 @@ The log4js.connectLogger supports the passing of an options object that can be u
 - log level
 - log format string or function (the same as the connect/express logger)
 - nolog expressions (represented as a string, regexp, or array)
+- status code rulesets
 
 For example:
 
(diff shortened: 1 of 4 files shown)


nomiddlename:
Great work here, thanks. Always nice to see the tests and the documentation. I'll push this out in the next release (4.0.0), but it could be a week or two before that happens.

blakambr:
Thanks for the speedy merge @nomiddlename ! And no worries. Enjoy the xmas break!

PR #860, opened 2019-04-24
fix(layout): Use level as a string for node v12

### LINKED ISSUE #859: console and stdout appenders start printing out whole Level object in node v12
looks like this:
[2019-04-24T08:47:32.597] [Level { level: 20000, levelStr: 'INFO', colour: 'green' }]

bjornstar:
Fixes #859

I would add tests for node v12, but there are a number of unrelated failures. I'll take a stab at getting the tests to pass under node v12 on a different PR. This one can solve the immediate problem that will stop log4js users from upgrading to node v12.

### DIFF
--- a/lib/layouts.js
+++ b/lib/layouts.js
@@ -45,7 +45,7 @@ function timestampLevelAndCategory(loggingEvent, colour) {
     util.format(
       '[%s] [%s] %s - ',
       dateFormat.asString(loggingEvent.startTime),
-      loggingEvent.level,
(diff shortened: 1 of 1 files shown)


nomiddlename:
Thanks for this - looks like there's quite a few things that break the tests in node 12 (a lot of them seem related to using sandboxed-module though).

nomiddlename:
Published to NPM in log4js@4.1.1

PR #862, opened 2019-04-25
feat(connectLogger): add response to context

leak4mk0:
Proposing a way to use a response of express in custom layouts or appenders.
In this version, `connectLogger` adds a response of express to context if `context` flag is set to `true`, and we can use it in custom layouts or appenders.


nomiddlename:
Thanks for this. The Travis CI failure is unrelated to your changes.

nomiddlename:
Published to NPM in version 4.2.0

PR #863, opened 2019-04-25
feat: add appender and level inheritance

pharapiak:
These changes:
- use a category's name to infer inheritance from another category
- merges in unique appenders from parent categories, recursively
- if no 'level' is defined for a category, it is copied from the nearest parent 
- fills in intermediate parent categories that are not explicitly declared in the config (E.g. given  catA and catA.catB.catC, catA.catB will be created if it doesn't exist. 
- added support for boolean category.inherit to disable this, for backwards compatibility.
- added addPreProcessingListener to config so that this inheritance is applied globally.
- updated terms.md.

See configuration-inheritance-test.js for examples.

### DIFF
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -59,8 +59,6 @@ function configure(configurationFileOrObject) {
   }
   debug(`Configuration is ${configObject}`);
 
-  categories.addInheritedConfig(configObject); // copy config from parent to child categories
-
   configuration.configure(deepClone(configObject));
 
   clustering.onMessage(sendLogEventToAppender);
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -59,6 +59,8 @@ function configure(configurationFileOrObject) {
   }
   debug(`Configuration is ${configObject}`);
 
+  categories.addInheritedConfig(configObject); // copy config from parent to child categories
+
   configuration.configure(deepClone(configObject));
 
   clustering.onMessage(sendLogEventToAppender);
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ _site
 Gemfile.lock
 Dockerfile
 docker-compose.yml
+
+#personal config
+.env
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # log4js-node changelog
 
+## 4.1.1
+* [Add missing types for addLevels](https://github.com/log4js-node/log4js-node/pull/867) - thanks [@Ivkaa](https://github.com/Ivkaa)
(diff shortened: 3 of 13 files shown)


lib/log4js.js
```diff
@@ -59,6 +59,8 @@ function configure(configurationFileOrObject) {
   }
   debug(`Configuration is ${configObject}`);
 
+  categories.addInheritedConfig(configObject); // copy config from parent to child categories
```
nomiddlename: I think this is going to give you problems when users supply read-only config objects (i.e. they've used `Object.freeze` or something similar). I think the `node-config` package does this by default. Your `addInheritedConfig` function modifies the original configObject. That's why the other modules add themselves as config listeners (there's code in the categories module that does this too) - they'll get called after the config object has been cloned, like below.

nomiddlename:
The travis CI failure is unrelated to your changes - I'll see if I can work out what's wrong with that test.

lib/log4js.js
pharapiak: Ah, I wasn't aware of read only config objects.

I found that I had to put some inheritance 'fixes' into the configuration validation, otherwise it rejects categories that have no directly configured appenders.  I was originally just adding inheritance to the cloned config that gets passed to validation, before I realized that left the original config without any inheritance.

Would it make sense to clone the incoming readonly config, and use that clone thereafter?  ie: apply inheritance, validate it, and use it for subsequent getLogger calls.

Another approach I considered was augmenting the categories object, with a getter method for appenders and level that considered inheritance, but I though it would be more efficient to just collapse inherited 'stuff' to sub-categories at startup.

nomiddlename: Yep, the steps should be the same as on the master branch for log4js.js:
```javascript
// log4js.js: line 62
configuration.configure(deepClone(configObject));
```
Then in configuration.js we add an extra set of listeners (`addPreProcessingListeners`?), and call those functions with the cloned configObject. They can modify the config before it gets validated. So categories.js would add three config listeners - one for pre-processing (apply inheritance), one for the validation rules, and one for the setup.

pharapiak:
Working on those requested changes.  The suggested preProcessingListeners approach seems to work cleanly.  Now reworking unit tests to verify the results.

pharapiak:
I added requested changes: applying inheritance through preProcessingListener.
I think this PR is ready to go.

nomiddlename:
Published to NPM in version 4.2.0

pharapiak:
Great!

BTW, this was a milestone for me:  my first-ever open source contribution.

nomiddlename:
Congratulations, and thanks again for your help.

PR #868, opened 2019-05-02
feat: Add request body token

Simmyoungsub:
I've added a body token to the connector-logger.

### DIFF
--- a/lib/connect-logger.js
+++ b/lib/connect-logger.js
@@ -114,7 +114,7 @@ function assembleTokens(req, res, customTokens) {
   defaultTokens.push({
     token: ':body',
     replacement: getBody(req)
-  })
+  });
 
   return arrayUniqueTokens(customTokens.concat(defaultTokens));
 }
--- a/lib/connect-logger.js
+++ b/lib/connect-logger.js
@@ -7,9 +7,7 @@ const levels = require('./levels');
 const DEFAULT_FORMAT = ':remote-addr - -'
   + ' ":method :url HTTP/:http-version"'
   + ' :status :content-length ":referrer"'
-  + ' ":user-agent"'
-  + ' ":body"';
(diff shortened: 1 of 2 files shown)


nomiddlename:
Thanks for this - could you add a test that covers it, please?

Simmyoungsub:
Branch was updated. 
Please check it~

lib/connect-logger.js
```diff
@@ -7,7 +7,9 @@ const levels = require('./levels');
 const DEFAULT_FORMAT = ':remote-addr - -'
   + ' ":method :url HTTP/:http-version"'
   + ' :status :content-length ":referrer"'
-  + ' ":user-agent"';
+  + ' ":user-agent"'
+  + ' ":body"';
```
nomiddlename: I wouldn't add body to the default log format - it is unusual for someone to want to log a request body in their http access logs. The default format is roughly equal to the standard Apache log format, so I don't want to change that.

Instead add something to the docs explaining how to add the body token to a custom format.

Simmyoungsub: As you know,
The custom token's replacement method can not access the request object.
related #717

Simmyoungsub:
So, I was add function what custom object had access to  request object

lib/connect-logger.js
Simmyoungsub: Please let me know if there is a way to access the request object.

nomiddlename: I didn't mean add "body" as a custom token, it can be one of the standard tokens but not part of the default format. So, take `:body` out of `DEFAULT_FORMAT`, but if someone wants to include the body in their log output then they can provide their own format string that includes `:body`.

Simmyoungsub: Yeah Yeah!! You're right!
I have an idea to solve this issue
I will add something to doc explaining how to use the "request body" if someone use a custom format.

I will regist another PR that i modify doc.

Issue #874, opened 2019-05-09
typescript docs not working


WeihanLi:
As the docs from [readme](https://github.com/log4js-node/log4js-node#typescript), it should be 

``` typescript
import { configure, getLogger } from './log4js';
configure('./filename');
const logger = getLogger();
// ...
```

but that's not working for me, I use the code like belows for instead:

``` typescript
import * as log4js from "log4js";
log4js.config({
  appenders: {
    file: { type: 'file', filename: './logs.log' },
    console: { type: 'console' }
  },
  categories: {
    default: { appenders: ['file', 'console'], level: 'debug' }
  }
});
let logger= log4js.getLogger("main");
```

and I'm using log4js version  4.1.0

leak4mk0:
Hi,
I think there is a error in the document.
We should import 'log4js' instead of './log4js', normally.
Could you try the following code?
```typescript
import { configure, getLogger } from 'log4js'; // <--- valid module name
configure('./filename');
const logger = getLogger();
// ...
```

WeihanLi:
I've tried `import { configure, getLogger } from 'log4js';` before but get an error like can not find module `log4js`, I think maybe there's something wrong with the types definition

WeihanLi:
I've tried a new project with latest log4js, it works greatly, maybe because I've had a `@types/log4js` installed before?

leak4mk0:
It can be possible because `@types/log4js` has been deprecated.
Can I see an error and npm version of `@types/log4js`?

WeihanLi:
I'm using the latest `@types/log4js` seemed to be `2.3.5`, can not reproduce the error for now, sorry to disturb...

Simmyoungsub:
"log4js.config" method is exist?

WeihanLi:
maybe like that, I forgot it... it's couple of days ago...

leak4mk0:
I'm glad that the problem was solved.
I'll close this issue after when my PR is merged.

WeihanLi:
OK, Thanks @leak4mk0

nomiddlename:
Docs fixed by @leak4mk0 - thanks!

leak4mk0:
I didn't understand who can close an issue.
Thank you for closing this issue, @nomiddlename

PR #877, opened 2019-05-12
fix(doc): name of required/imported module in document.

leak4mk0:
I saw #874 and found some errors related module name in document.
So, I'll fix them.


leak4mk0:
close #874

nomiddlename:
Thanks for picking this up.

PR #879, opened 2019-05-15
add line number support

victor0801x:
- logger add `enableCallStack(boolean)`
- more layout support
  - `%f` fileName
  - `%l` lineNumber
  - `%o` columnNumber
  - `%s` call stack

### DIFF
--- a/lib/log4js.js
+++ b/lib/log4js.js
@@ -33,16 +33,6 @@ const clustering = require('./clustering');
 const connectLogger = require('./connect-logger');
 
 let enabled = false;
-let gCallStackEnabled = false;
-const loggers = new Map();
-
-function useCallStack(bool = true) {
-  gCallStackEnabled = (bool === true);
-
-  loggers.forEach(logger => logger.enableCallStack(gCallStackEnabled));
-  // eslint-disable-next-line no-use-before-define
-  return log4js;
-}
 
 function sendLogEventToAppender(logEvent) {
   if (!enabled) return;
@@ -133,15 +123,7 @@ function getLogger(category) {
       categories: { default: { appenders: ['out'], level: 'OFF' } }
     });
   }
-  category = category || 'default';
-  let logger = loggers.get(category);
-  if (!logger) {
-    logger = (new Logger(category)).enableCallStack(
-      categories.isCallStackEnable(category) || gCallStackEnabled
-    );
-    loggers.set(category, logger);
-  }
-  return logger;
+  return new Logger(category || 'default');
 }
 
 
@@ -159,7 +141,6 @@ const log4js = {
   connectLogger,
   levels,
   addLayout: layouts.addLayout,
-  useCallStack,
 };
 
 module.exports = log4js;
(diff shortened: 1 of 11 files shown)


lib/logger.js
```diff
@@ -67,6 +91,15 @@ class Logger {
   clearContext() {
     this.context = {};
   }
+
+  enabelCallStack(bool = true) {
```
nomiddlename: Spelling mistake here - should be `enableCallStack`

victor0801x:
I prefer second XD
ps: I try to auto detect appender's layout.patter `/%[f|l|o|s]/.test() === true` then enableCallStack

victor0801x:
@nomiddlename I saw [in /lib/log4js:126](https://github.com/log4js-node/log4js-node/blob/master/lib/log4js.js#L126)
even the category is same, still return new Logger(), is this correct ?

victor0801x:
@nomiddlename 
last commit
- add a loggers `Map` to store the same category's logger => try to reduce LoggerEvent count
- log4js add `useCallStack(boolean)` for enable/disable all loggers' callstack

lib/logger.js
```diff
@@ -67,6 +90,15 @@ class Logger {
   clearContext() {
     this.context = {};
   }
+
+  enableCallStack(bool = true) {
```
nomiddlename: `enableCallStack` with an argument seems confusing to me. This might be better as a property of the Logger class, with `get` and `set` functions which validate the boolean value (similar to how the `level` property is defined on Logger). This would also avoid having to cache Loggers by setting the `useCallStack` value at the category level (like how the log level is set at the category level).

lib/log4js.js
```diff
@@ -140,7 +158,8 @@ const log4js = {
   shutdown,
   connectLogger,
   levels,
-  addLayout: layouts.addLayout
+  addLayout: layouts.addLayout,
+  useCallStack,
```
nomiddlename: I don't think you need the global useCallStack - keeping it at the category level seems cleaner and simpler.

lib/logger.js
victor0801x: But, that way will influence after new Logger(), not respect to the configuration

nomiddlename: Take a look at how the `level` property is defined - it always looks up the level in the category configuration, so it applies to all loggers.

victor0801x: yup, applies to all same category loggers.

lib/log4js.js
```diff
@@ -33,6 +33,16 @@ const clustering = require('./clustering');
 const connectLogger = require('./connect-logger');
 
 let enabled = false;
+let gCallStackEnabled = false;
+const loggers = new Map();
```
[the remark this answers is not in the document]
victor0801x: fixed

lib/log4js.js
victor0801x: fixed

nomiddlename:
Published to NPM in 4.3.0

lileilei:
image
Missing fields need to be supplemented

s97712:
Would it be better to implement it with babel plugin?

rfox12:
It looks like these fields didn't make it into the type description for `LoggingEvent` (see `log4js.d.ts`)

Issue #816, opened 2018-12-04
Problem using webpack with logLevelFilter


hanaeddy:
3.0.6: got this error after webpacking code when using logLevelFilter appender :
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined

I only succeed in fixing it after pre-loading logLevelFilter appender in lib/appenders/index.js:16:
coreAppenders.set('logLevelFilter', require('./logLevelFilter'));

nomiddlename:
Thanks for raising this - I think this has been raised in another issue as well. It should be a simple fix like you've already done, so I'll include it in the 4.0.0 release.

ganlyun:
@nomiddlename same issue in the 4.2.0 release with webpack v4.31.0, it seems not be fixed??

nomiddlename:
Aargh. You're right - looks like what happened was @hanaeddy closed the issue thinking I'd already made the change, and then I totally forgot about it because the issue was closed.

nomiddlename:
Fix published to NPM in version 4.3.0 (for real, this time)

guntram:
I get this error for the config:
`{
      appenders: {
        logstash: {
          type: '@log4js-node/logstashudp',
          host: this.logstashUrl,
          port: this.logstashPort,
          extraDataProvider: (loggingEvent) => ({
            host: {
              osPlatform: osPlatform,
            },
            tags: this.logstashTags,
            log: {
              level: this.logstashMinimumLevel,
            },
            panelVersion: panelVersion,
          }),
        },
      },
      categories: {
        default: {
          appenders: ['logstash'],
          level: this.logstashMinimumLevel,
        },
      },
    }`

using log4js@6.3.0 and @log4js-node/logstashudp@1.2.1 in an angular11 app which runs inside chromium embedded framework.
This setup is working fine in an older angularJs app...

PR #882, opened 2019-05-20
fix: loglevelfilter was missing from core appenders

nomiddlename:
This is a fix for #816 that got completely forgotten about as far as I can tell. Sorry.


nomiddlename:
Published to NPM in 4.3.0

Issue #887, opened 2019-05-27
DateFile ignores maxLogSize


nomiddlename:
The log4js dateFile docs say that any file appender option can be used with the dateFile appender, but maxLogSize (file appender config) does not work - this is because the streamroller library is expecting maxSize.

nomiddlename:
Fix published to NPM in version 4.3.1

PR #889, opened 2019-05-28
fix(#887): map maxLogSize to maxSize

nomiddlename:


nomiddlename:
Published to NPM in version 4.3.1

PR #900, opened 2019-06-19
chore: updated tap to 14, other deps to latest

nomiddlename:


nomiddlename:
Looks like tap v14 breaks the travis build on node 6.

Issue #888, opened 2019-05-28
Possible to truncate filename in logger


mLjH:
Is there anyway to truncate the whole absolute path when logging?

Right now, my layout for logging is 
```
"layout": {
   "type": "pattern",
   "pattern": "[%d{yyyy-MM-ddThh:mm:ss.SSS}] [%p] %c %f %l - %m"
},
```
But the filename option is printing out the whole absolute path to the file. I would like to have possibly the option to see either
**filename.js** OR **.../folder/filename.js**

nomiddlename:
There's no support for parsing file paths in pattern layout. You could add your own custom layout, or I'd be happy to see a pull request that added in formats for the filename field (similar to how you can use `%c{2}` to only output two levels of category hierarchies. Alternatively a pull request that added right-truncation in addition to the current left truncation would be good.

Simmyoungsub:
How about this?
Use the latest version.

```
const path = require('path');
const log4js = require('log4js');
log4js.configure({
    appenders: {
        out: {
            type: 'stdout',
            layout: {
                type: 'pattern',
                pattern: '%d %p %c %m (%x{fileName}) %n',
                tokens: {
                    fileName: (logEvent) => {
                        return logEvent.fileName.split(path.sep).slice(-1)[0] || '';
                    }
                }
            }
        }
    },
    categories: { default: { appenders: ['out'], level: 'INFO' , enableCallStack: true} }
});
```

nomiddlename:
Published to npm in log4sj@4.5.0

PR #913, opened 2019-07-09
feat: add delimiter for fileName depth to %f layout patern

rommni:
solved #888 without using the token way, that i personnally found less clear.


nomiddlename:
Published to npm in log4sj@4.5.0

PR #914, opened 2019-07-09
feat: allow to change default way to parse callstack

rommni:
Add a function to the logger to edit the function parsing the callstack.

Use case example: Can be use when all the call to logger.log are made from a custom "debug" class (called from outside with debug.log, debug.warn, etc ... ), doing the interface between debuging an app and log4js, to allow more flexibility,  and you want to use layout patern to show the fileName where the debug.x is called.

You can this way create a custom function parsing callStack and removing lines concerning your "debug" class, and then give it to the logger.


rommni:
Sorry for the test forgot them and like i'm on window at work and can't run test localy properly i didn't saw it before, trying to make a test after eating ....

nomiddlename:
Published to npm in log4sj@4.5.0

PR #918, opened 2019-07-12
chore(deps): streamroller 1.0.5 -> 1.0.6

### LINKED ISSUE #911: Testing size-based rotation, doesn't seem to work as expected
I'm still running tests on log4js assessing it's viability as a replacement for the no longer maintained bunyan.

I just did a very very basic test on size-based rotation, and I'm not seeing the behaviour I'd expect.

Here's my test code:

```
const config = {
    appenders: {
        access: {type: 'file', filename: 'logs/test.log', maxLogSize: 5000, backups: 4}
    },
    categories: {default: {appenders: ['access'], level: 'debug'}}
};

log4js.configure(config);

const logger = log4js.getLogger('default');

logger.info('hello world 1');
logger.error(new Error('bad thing 1'));
logger.info('hello world 2');
logger.error(new Error('bad thing 2'));
```

This file is saved as `logging.js` and I'm hammering `node logging.js` locally on my macbook pro. I'm expecting there to quickly be 10 files, but there are never more than 2. 

If I duplicate my log lines at the bottom so much more log content is output, I can get 10 files to be created, but never from consistently hitting the script.

I thought it might be related to this bug, because outputting more content in one stream/session creates more files: https://github.com/log4js-node/streamroller/issues/34 - but it seems to be entirely to do with the number of backup files created, not the size of the individual files.

I'm looking for file size based rotation, not date based rotation, due the sheer volume of log files I'm going to be using this with - rotating all the log files across 1000s of apps at midnight causes performance problems 😬

nomiddlename:
This will fix #911 I think.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1648,7 +1648,7 @@
     "fs-extra": {
       "version": "7.0.1",
       "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-7.0.1.tgz",
-      "integrity": "sha1-TxicRKoSO4lfcigE9V6iPq3DSOk=",
+      "integrity": "sha512-YJDaCJZEnBmcbw13fvdAM9AwNOJwOzrE4pqMqBq5nFiEqXUqHwlK4B+3pUw6JNvfSPtX05xFHtYy/1ni01eGCw==",
       "requires": {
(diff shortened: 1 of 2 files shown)


nomiddlename:
Published to NPM in log4js@4.5.1

PR #938, opened 2019-09-08
[enhancement]Emmit pause event for node js when buffer is full

shayantabatabaee:
Hi,

I made some change in file.js module, so when the buffer is full the 'pause' event will emit, and after now we can listen to this event. When this event is emitted we can stop logging or maybe send 'server too busy' response and wait till the drain event emitted. With these changes logger will not stop and just an event emitted, and this is possible if someone want to continue logging.

### DIFF
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -15,6 +15,9 @@ function openTheStream(file, fileSize, numFiles, options) {
   stream.on('error', (err) => {
     console.error('log4js.fileAppender - Writing to file %s, error happened ', file, err); //eslint-disable-line
   });
+  stream.on('drain', () => {
+    process.emit("pause", false);
+  });
   return stream;
 }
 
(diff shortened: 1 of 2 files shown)


nomiddlename:
Thanks for this, could you add a test that covers the change please?

shayantabatabaee:
Yes, I’ll add the test as soon as possible.

shayantabatabaee:
@nomiddlename I've added test for this situation.

shayantabatabaee:
thanks @nomiddlename for response, your are right I'will add the namespace to event.
For the second problem that you mentioned this event just emit if user get buffer more than it's allowed and this is just for writing to file, no matter how many files to write. I think when the user receives this event can decide to stop logging to file but continue to send over the network or even write in console. I mean this event only emit while writing to files and we can add this note to documents

nomiddlename:
Yep, I agree - let's start with this, and see if it's an approach that works for people. I'll create a couple of extra github issues to add documentation and also make the same changes in the dateFile appender. Once those are done, I'll push it out to NPM.

nomiddlename:
To be clear, I don't expect you to work on those two extra issues - I just created them so that I can track what else needs to be done before I can release the change. If you want to work on them, that's awesome, but you don't have to. I'll work on them as I get time.

nomiddlename:
Published to npm in log4js@6.1.0

PR #939, opened 2019-09-09
fix: slack (incoming webhook) appender

jhfoo:
Slack messaging API changed. This PR interfaces with the incoming webhook API described here: https://api.slack.com/incoming-webhooks

Example: slack-hook-appender.js


nomiddlename:
Thanks for your work, but could you raise this pull request against the [slack appender project](https://github.com/log4js-node/slack), please? The slack appender does not belong in the core of the log4js project.

Issue #955, opened 2019-10-19
Pattern format for call stack using padding/truncation not working


lheberlie:
Thank you so much for enabling the call stack in the core branch.

**Problem**: I am having issues getting the padding and truncation working with `%f`, `%l`, and `%o`. When using truncation with `%l` and `%o`, there are fatal errors, however when using padding/truncation with `%f` there are no errors but it doesn't seem to be applying the padding/trunction.

Here is my `log4js.config.json` file.

```
{
  "appenders": {
    "console-appender": {
      "type": "console",
      "layout": {
        "type": "pattern",
        "pattern": "%[[%p]%] - %18.100f{2} | %7.12l:%7.12o - %[%m%]"
      }
    }
  },
  "categories": {
    "default": {
      "appenders": ["console-appender"],
      "enableCallStack": true,
      "level": "info"
    }
  }
}
```

**Error**:

```
/node_modules/log4js/lib/layouts.js:272
      return toTruncate.substring(0, len);
                        ^

TypeError: toTruncate.substring is not a function
    at truncate (/node_modules/log4js/lib/layouts.js:272:25)
    at truncateAndPad (/node_modules/log4js/lib/layouts.js:300:19)
    at /node_modules/log4js/lib/layouts.js:326:28
    at /node_modules/log4js/lib/appenders/console.js:6:16
    at /node_modules/log4js/lib/log4js.js:42:5
    at Array.forEach (<anonymous>)
    at sendLogEventToAppender (/node_modules/log4js/lib/log4js.js:41:21)
    at /node_modules/log4js/lib/clustering.js:23:26
    at Array.forEach (<anonymous>)
    at sendToListeners (/node_modules/log4js/lib/clustering.js:23:13)
    at Object.send (/node_modules/log4js/lib/clustering.js:89:7)
    at Logger._log (/node_modules/log4js/lib/logger.js:90:16)
    at Logger.log (/node_modules/log4js/lib/logger.js:73:12)
    at Logger.<computed> [as info] (/node_modules/log4js/lib/logger.js:124:10)
    at Object.<anonymous> (/bin/create-sample-code-thumbnails.js:108:10)
```

**Additional details**:

When I remove the truncation from `%l` and `%o` the code will work but it doesn't seem to be formatting correctly.

*Resulting output*

```
[INFO] - sample-code-helper/screenshot-helper.js | 631:16 - screenshot-helper - converting screenshots to thumbnails - end
[...]

PR #956, opened 2019-10-21
fix: truncation and padding for stacktraces

nomiddlename:
This fixes the errors found in #955, and also adds negative truncation (allowing you to truncate fields from the end of the string instead of the start).


nomiddlename:
Published to NPM in version 5.3.0

Issue #971, opened 2019-11-20
Inconsistent typings for shutdown function


aaron-edwards:
There seems to be an inconsistent typing of the shutdown function. 

The definition file specifies the shutdown function as follows 
```typescript
export function shutdown(cb?: (error: Error) => void): void | null;
```
however the function seems to require a cb function (See https://github.com/aaron-edwards/log4js-node/blob/master/lib/log4js.js#L106 & https://github.com/aaron-edwards/log4js-node/blob/master/lib/log4js.js#L112)

nomiddlename:
Fix published to npm in log4js@6.1.1

PR #972, opened 2019-11-20
fix: Adds guards for undefiend callback functions in shutdown method

aaron-edwards:
This fixes issue #971


nomiddlename:
Thanks for picking this up.

nomiddlename:

PR #984, opened 2020-01-24
fix(types): mark log level method in Logger class declaration.

techmunk:


nomiddlename:
Cool - thanks for spotting that and fixing it.

nomiddlename:

PR #985, opened 2020-01-28
Update Loading mechanism instruction #1

leonimurilo:
Update Loading mechanism instruction (number 1 in priority) to match the current behaviour of the library.


nomiddlename:
Hi, thanks for this - while I agree that the code does something slightly different to the docs, I don't think your version (or the original) one explains what it is doing which is checking for the appender in the bundled core library appenders.

PR #986, opened 2020-01-31
fix: Allow appenders to depend on not-yet-loaded appenders.

### LINKED ISSUE #746: TypeError: appender is not a function
The order in which the appenders are defined seems _very_ important. This gives some strange results:

This sample:

```ts
log4js.configure({
  appenders: {
     filteredConsole: { type: 'logLevelFilter', appender: 'console', level: 'info' },
     console: { type: 'stdout' }
  },
  categories: {
    default: { appenders: ['filteredConsole'], level: 'trace' }
  }
});
log4js.getLogger('demo').info('test');
```

Results in error: `TypeError: appender is not a function`. 

A solution would be to sort the appenders topologically. We could use something like [toposort](https://www.npmjs.com/package/toposort)

mvastola:
Close #746

### DIFF
--- a/lib/appenders/index.js
+++ b/lib/appenders/index.js
@@ -40,6 +40,22 @@ const loadAppenderModule = (type, config) => coreAppenders.get(type)
   || (require.main && tryLoading(path.join(path.dirname(require.main.filename), type), config))
   || tryLoading(path.join(process.cwd(), type), config);
 
+const appendersLoading = new Set();
+
+const getAppender = (name, config) => {
+  if (appenders.has(name)) return appenders.get(name);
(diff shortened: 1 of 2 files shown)


mvastola:
@nomiddlename test is added (and passing).

mvastola:
Actually, just spotted a bug. Give me a min.

mvastola:
Ok. Now it should be good.

nomiddlename:
Thanks for this - published to NPM in  log4js@6.1.2

mvastola:
@nomiddlename Thanks!!

PR #996, opened 2020-04-19
chore: update streamroller to 2.2.4

nomiddlename:
Streamroller 2.2.4 contains a fix for incorrect filename matching when cleaning up old files.


nomiddlename:
Published to npm in log4js@6.2.1

PR #1001, opened 2020-05-01
Added support for removing embedded ANSI color sequence for file appender

BlueCocoa:
It's quite common for developers to use embedded color sequence like `\x1b[33m` in the log string to make the output look beautiful. 

However, if the output is also set to append to some file, the embedded color sequence will write to the given file as is, which may not be expected in some cases. 

Thus I added one more option for file appender, which is `removeColor`. It's a boolean value that controls whether if the appender should detect and remove embedded color sequence before writing to file.

When this option is set to `false`, everything will be written to file as is. For example,

```javascript
log4js.configure({
    appenders: {
        Core: { type: "file", filename: "app.log", removeColor: false },
        console: { type: "console" }
    },
    categories: {
        bot: { appenders: ["console", "Core"], level: "trace" },
        default: { appenders: ["console"], level: "trace" }
    }
})
const botLogger = log4js.getLogger("bot")
botLogger.debug("\x1b[33mHello\x1b[0m")
```

In the `app.log`, it will be `[33mHello[0m`, `\x1b` is invisible.

When this option is set to `true`, then embedded color sequence will be removed before writing to file.

```javascript
log4js.configure({
    appenders: {
        Core: { type: "file", filename: "app.log", removeColor: true },
        console: { type: "console" }
    },
    categories: {
        bot: { appenders: ["console", "Core"], level: "trace" },
        default: { appenders: ["console"], level: "trace" }
    }
})
const botLogger = log4js.getLogger("bot")
botLogger.debug("\x1b[33mHello\x1b[0m")
```

Now in the `app.log`, it will be `Hello` only.

Signed-off-by: Cocoa <0xbbc@0xbbc.com>

### DIFF
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -55,12 +55,8 @@ function fileAppender(file, layout, logSize, numBackups, options, timezoneOffset
   const app = function (loggingEvent) {
     if (options.removeColor === true) {
       // eslint-disable-next-line no-control-regex
-      const regex = new RegExp("\x1b[[0-9;]*m", "g");
-      for (let i = 0; i < loggingEvent.data.length; i += 1) {
-        let d = loggingEvent.data[i];
-        d = d.replace(regex, '');
-        loggingEvent.data[i] = d;
-      }
+      const regex = /\x1b[[0-9;]*m/g;
+      loggingEvent.data = loggingEvent.data.map(d => d.replace(regex, ''))
     }
     if (!writer.write(layout(loggingEvent, timezoneOffset) + eol, "utf8")) {
       process.emit('log4js:pause', true);
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -58,7 +58,7 @@ function fileAppender(file, layout, logSize, numBackups, options, timezoneOffset
       const regex = /\x1b[[0-9;]*m/g;
       loggingEvent.data = loggingEvent.data.map(d => {
         if (typeof d === 'string') return d.replace(regex, '')
-        else return d
+        return d
       })
     }
     if (!writer.write(layout(loggingEvent, timezoneOffset) + eol, "utf8")) {
--- a/lib/appenders/file.js
+++ b/lib/appenders/file.js
@@ -53,6 +53,15 @@ function fileAppender(file, layout, logSize, numBackups, options, timezoneOffset
   let writer = openTheStream(file, logSize, numBackups, options);
 
   const app = function (loggingEvent) {
+    if (options.removeColor === true) {
+      let data = loggingEvent.data
+      for (let i = 0; i < data.length; i++) {
+        let d = data[i]
+        d = d.replace(/\x1b\[[0-9]{1,2}m/g, '')
+        data[i] = d
+      }
+      loggingEvent.data = data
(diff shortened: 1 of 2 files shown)


rnd-debug:
@BlueCocoa I am not a maintainer, but thank for the idea! Can you fix those ESLint errors?
image
There is a couple of other cases to consider, like three-digits color codes and codes separated by a semicolon (see [https://en.wikipedia.org/wiki/ANSI_escape_code](url) and [https://superuser.com/questions/380772/removing-ansi-color-codes-from-text-stream](url))
Maybe something like this will be more accurate:
`const regex = new RegExp("\x1b[[0-9;]*m", "g");`

BlueCocoa:
> @BlueCocoa I am not a maintainer, but thank for the idea! Can you fix those ESLint errors?
> image
> 
> There is a couple of other cases to consider, like three-digits color codes and codes separated by a semicolon (see [https://en.wikipedia.org/wiki/ANSI_escape_code](url) and [https://superuser.com/questions/380772/removing-ansi-color-codes-from-text-stream](url))
> Maybe something like this will be more accurate:
> `const regex = new RegExp("\x1b[[0-9;]*m", "g");`

Hi @rnd-debug, thanks for your kindly reply and the hint on ANSI color codes. I've made corresponding changes in the second commit. I think that should fix all the issues that ESLint complained.

BlueCocoa:
According to the [ESLint docs](https://eslint.org/docs/rules/no-control-regex), perhaps we should turn off `no-control-regex` rule for this file?

lib/appenders/file.js
```diff
@@ -53,6 +53,15 @@ function fileAppender(file, layout, logSize, numBackups, options, timezoneOffset
   let writer = openTheStream(file, logSize, numBackups, options);
 
   const app = function (loggingEvent) {
+    if (options.removeColor === true) {
+      // eslint-disable-next-line no-control-regex
+      const regex = new RegExp("\x1b[[0-9;]*m", "g");
```
nomiddlename, line 58: Use the regexp literal (e.g. `regex = /\x1b[..../g`) instead of `new RegExp` constructor when the pattern won't change. The literal gets evaluated once, whereas the constructor will create a new one every time.

lib/appenders/file.js
```diff
@@ -53,6 +53,15 @@ function fileAppender(file, layout, logSize, numBackups, options, timezoneOffset
...
+      for (let i = 0; i < loggingEvent.data.length; i += 1) {
```
nomiddlename, line 59: It would be better to use a map here instead of the loop:
`loggingEvent.data = loggingEvent.data.map(d => d.replace(regex, ''))`

BlueCocoa:
Hi @nomiddlename, much thanks to your suggestions, corresponding changes has been made and pushed :-D

nomiddlename:
Sorry @BlueCocoa - I just thought of a problem with this. The `loggingEvent.data` array can contain things that are not strings, for example if you do this:
```javascript
const thing = { a: 'value' };
logger.debug('this is the thing: ', thing);
```
I think your code will throw an error in this case.

nomiddlename:
Fix published in log4js@6.3.0

Issue #1000, opened 2020-04-29
File appenders try to open files even though they do not exist in a category


thernstig:
I have a setup where I've configured a file appender on a logging config object, but I never added that file appender to any category.

When log4js loads the config, it then prints:

```sh
log4js.fileAppender - Writing to file /path/to/server.log, error happened  [Error: EACCES: permission denied, open '/path/to/server.log'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '/var/opt/ericsson/cnom/log/eric-cnom-server.log'
}
```

Is this intentional? Or could the appenders avoid trying to open the file until they have actually been added to a category?

nomiddlename:
Fix published in log4js@6.3.0

PR #1002, opened 2020-05-03
Donot create appender if not used

rnd-debug:
Related to #1000 . The fix will check if an appender exists in a category before trying to create it.

### DIFF
--- a/lib/appenders/index.js
+++ b/lib/appenders/index.js
@@ -99,7 +99,7 @@ const setup = (config) => {
   });
 };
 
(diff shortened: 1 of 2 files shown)


rnd-debug:
Interesting: fails on `test/tap/server-test.js` with :
> ✖ connect ECONNREFUSED 127.0.0.1:5678
Any idea why?

nomiddlename:
Fix published in log4js@6.3.0 - the ECONNREFUSED error was due to the `tcp-server` appender not being started, it just listens for tcp messages and forwards them to other appenders so it does not need a category assigned. I hacked in a dodgy fix for it.

rnd-debug:
Thank you for the feedback!

PR #1003, opened 2020-05-03
Describe categories inheritance

rnd-debug:
Noticed that the categories inheritance / hierarchy principle was described only in:
- [https://github.com/log4js-node/log4js-node/blob/master/docs/terms.md](url)
- and in the test files [https://github.com/log4js-node/log4js-node/blob/master/test/tap/configuration-inheritance-test.js](url)

Maybe it is worth to provide additional info on the appenders page and/or on a separate dedicated page.


nomiddlename:
Awesome, thanks for doing this.

PR #1004, opened 2020-05-04
style: conform @param to official style

wataash:
ref. https://jsdoc.app/tags-param.html


nomiddlename:
Thanks for doing this!

PR #1006, opened 2020-05-06
Access to 'category' field of logger objects in TS

rtvd:
Sometimes it can be useful to know the category of the logger.
The category can be read from pure JavaScript code but TypeScript code cannot read it as the field is not declared.

This change adds the field to the TypeScript declaration. The access to it is read-only as changing the category of logger probably makes no sense.


nomiddlename:
Thanks a lot for this PR.

nomiddlename:
Fix published in log4js@6.3.0

Issue #1007, opened 2020-05-07
Add udp appender


iassasin:
Hello! I've created appender 'udp', like 'tcp'. Can it be added to list of optional appenders?
https://www.npmjs.com/package/log4js-udp-appender

nomiddlename:
Awesome - just create a PR to modify the documentation.

iassasin:
Merged, thanks!

PR #1018, opened 2020-05-24
Add section in documentation for links to other appenders

iassasin:
Relates to #1007.
It seems not good to add to optional appenders, so I've created new section for it.


nomiddlename:
Thanks!

Issue #968, opened 2019-11-20
TypeError: Cannot read property 'native' of undefined


Eric8511:
I use this module in a web project with node v10.16.0, npm 6.9.0, vue 2.9.6, and got an error when start the website:
index.js?8373:107 Uncaught TypeError: Cannot read property 'native' of undefined
    at eval (index.js?8373:107)
    at Object../node_modules/streamroller/node_modules/fs-extra/lib/fs/index.js (app.js:16953)
    at __webpack_require__ (app.js:767)
    at fn (app.js:130)
    at eval (index.js?215d:6)
    at Object../node_modules/streamroller/node_modules/fs-extra/lib/index.js (app.js:16965)
    at __webpack_require__ (app.js:767)
    at fn (app.js:130)
    at Object.eval (RollingFileWriteStream.js?9f14:2)
    at eval (RollingFileWriteStream.js:269)

georgezouq:
same problem, any solution?

nomiddlename:
Looks like you're trying to use webpack to bundle up code to use in a browser, but you're also configuring a rolling file stream - file streams won't work in the browser.

bryanSME:
@nomiddlename is there no way to use this without the rolling file stream?

nomiddlename:
I don't think that warning would come from using a file appender, because the file appenders are listed in the core appenders - which avoids the dependency-is-an-expression problem. Could you give me a bit more detail what version of log4js you're using, and what your config looks like, please?

bryanSME:
Trying to make this as simple as possible, I've tried no config. Just adding the following causes the issue.  Running this in node has no issue for me. Running it on any code that hits the browser is where the warnings/errors start.  **Should I open a new ticket?**

version: 5.3.0 and 6.1.0

Here is the simplest code that causes the warning and failed compilation.
```
import { getLogger } from 'log4js';
getLogger(); 
```

WARNING in ./node_modules/log4js/lib/appenders/index.js 25:11-30
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/log4js/lib/log4js.js
 @ ./src/...../........
 @ ./src/index.jsx
Child html-webpack-plugin for "index.html":
         Asset     Size  Chunks  Chunk Names
    index.html  537 KiB       1  
    Entrypoint undefined = index.html
    [./node_modules/html-webpack-plugin/lib/loader.js!./src/index.ejs] 2.69 KiB {1}
[...]

PR #1023, opened 2020-05-27
fix: log4js-node/log4js-node#968

aellerton:
This addresses #968 by allowing certain `require`'s to run only in node, not in the browser.

All tests pass as before. However, I didn't add new tests specifically for this, which I
acknowledge isn't great, but I'm not sure how to write tests for "this works when
building with webpack" beyond adding a whole webpack project to the tests, and
grepping output. There might be a great way, but I'm not sure what it is.

I also found it useful to add this in the `package.json` of the project using log4js,
which I'm adding as a note here only in case it is useful for others facing this problem.

```
"browser": {
  "fs": false,
  "cluster": false
},
```
The initial PR build failed on Windows for unknown reasons; hopefully another build attempt will resolve that.

### DIFF
--- a/lib/appenders/index.js
+++ b/lib/appenders/index.js
@@ -6,6 +6,11 @@ const levels = require('../levels');
 const layouts = require('../layouts');
 const adapters = require('./adapters');
 
+const isNode =
+  typeof process !== 'undefined' &&
+  process.versions != null &&
+  process.versions.node != null;
+
 // pre-load the core appenders so that webpack can find them
 const coreAppenders = new Map();
 coreAppenders.set('console', require('./console'));
@@ -14,9 +19,9 @@ coreAppenders.set('stderr', require('./stderr'));
 coreAppenders.set('logLevelFilter', require('./logLevelFilter'));
 coreAppenders.set('categoryFilter', require('./categoryFilter'));
 coreAppenders.set('noLogFilter', require('./noLogFilter'));
-coreAppenders.set('file', require('./file'));
-coreAppenders.set('dateFile', require('./dateFile'));
-coreAppenders.set('fileSync', require('./fileSync'));
+coreAppenders.set('file', isNode ? require('./file') : {});
+coreAppenders.set('dateFile', isNode ? require('./dateFile') : {});
+coreAppenders.set('fileSync', isNode ? require('./fileSync'): {});
 
 const appenders = new Map();
 
--- a/lib/appenders/index.js
+++ b/lib/appenders/index.js
@@ -6,6 +6,7 @@ const levels = require('../levels');
 const layouts = require('../layouts');
 const adapters = require('./adapters');
 
+// Webpack needs some appenders to be screened out.
 const isNode =
   typeof process !== 'undefined' &&
   process.versions != null &&
--- a/lib/clustering.js
+++ b/lib/clustering.js
@@ -2,13 +2,20 @@ const debug = require("debug")("log4js:clustering");
 const LoggingEvent = require("./LoggingEvent");
 const configuration = require("./configuration");
 
-let disabled = false;
+const isNode =
+  typeof process !== 'undefined' &&
+  process.versions != null &&
+  process.versions.node != null;
+
+let disabled = !isNode;
 let cluster = null;
-try {
-  cluster = require("cluster"); //eslint-disable-line
-} catch (e) {
-  debug("cluster module not present");
-  disabled = true;
+if (!disabled) {
+  try {
+    cluster = require("cluster"); //eslint-disable-line
+  } catch (e) {
+    debug("cluster module not present");
+    disabled = true;
+  }
 }
 
 const listeners = [];


lib/appenders/index.js
```diff
@@ -6,6 +6,11 @@ const levels = require('../levels');
 const layouts = require('../layouts');
 const adapters = require('./adapters');
 
+const isNode =
+  typeof process !== 'undefined' &&
+  process.versions != null &&
+  process.versions.node != null;
+
```
Download, line 14: Just a nitpick, but why check against `null`?
I would expect:

```js
const isNode = typeof process !== 'undefined' && process.versions && process.versions.node
```
```

File: lib/appenders/index.js [hunk 2]
```diff
@@ -14,9 +19,9 @@ coreAppenders.set('stderr', require('./stderr'));
 coreAppenders.set('logLevelFilter', require('./logLevelFilter'));
 coreAppenders.set('categoryFilter', require('./categoryFilter'));
 coreAppenders.set('noLogFilter', require('./noLogFilter'));
-coreAppenders.set('file', require('./file'));
-coreAppenders.set('dateFile', require('./dateFile'));
-coreAppenders.set('fileSync', require('./fileSync'));
+coreAppenders.set('file', isNode ? require('./file') : {});
+coreAppenders.set('dateFile', isNode ? require('./dateFile') : {});
+coreAppenders.set('fileSync', isNode ? require('./fileSync'): {});
```
Download, line 25: I wonder whether there is a better way to do this...

There is a field in `package.json` called `browser` that might be better suited for this use case.
The way it works is this: You specify which file(s) to use instead of the default one(s), when running on the browser.
For example, we could make a new file called `core-appenders.js` and place the lines creating the `coreAppenders` map and populating it with the core appenders in that file. We could then make a second file called `core-appenders-browser.js` and place the equivalent code for the browser in that file. Then, in `package.json` we would write:

```json
{
  "browser": {
    "./lib/appenders/core-appenders.js": "./lib/appenders/core-appenders-browser.js"
  }
}
```

Then, Browserify, Webpack and Rollup et all will know to replace the code in `core-appenders.js` with the code in `core-appenders-browser.js` when building the browser version.

An example where this is used is the popular logging package `debug`:
https://github.com/visionmedia/debug/blob/4.2.0/package.json#L55

There are two forms for the `browser` field:

- Object with node file names as keys and browser filenames as values
- File name of entry point to use instead of the one listed in the `main` field.

lib/appenders/index.js
aellerton: Good call. I should have seen this.

lib/appenders/index.js
aellerton: I hadn't realised this - nicely spotted, and clearly explained! Only thing that comes to mind is a possible ordering issue, or that there would be code duplication and one of those files (node or browser) could get left behind. I'm not sure i that's a higher price or the runtime cost; given the runtime cost is (as far as I can see) once off, I think it might be better to leave it as-is. More thought required!

aellerton:
> I took the liberty to review your code, even though I'm not in any way affiliated with this project :)
> Hope it's helpful.

So happy to get your code review! Thanks for your time!

lib/appenders/index.js
Download: An important thing to understand about this line of code:

```js
coreAppenders.set('file', isNode ? require('./file') : {});
```

Even though the `isNode` check will prevent the file appender from being added to the `coreAppenders` map on the browser, it actually still gets bundled into the browser bundle. The reason is that, even though Webpack is pretty clever, it's not actually smart enough to figure out that `isNode` will always be `false` on browsers. Basically what Webpack does is find all required files in a static analysis phase and make sure each and every single file that *might* be required at runtime gets packaged into the bundle. 

Using the `browser` field to inform Webpack and other bundlers about the fact that the file appender is not used on browsers is the accepted/standard way to do this.

> there would be code duplication

That depends entirely on how you structure things. Make sure you place only the code that differs between Node and browsers in separate files and then use the `browser` field to map the browser versions of those files. No code needs to be duplicated.

CuddlySheep:
I had the problem executing my project in the browser due to a dependency on a package that itself has a dependency on log4js.
I've tested this bugfix in my project and can confirm that it works.

beorn:
Any update on this? :-|

aellerton:
Yeah sorry, I've had a series of release deadlines that have stopped me from tackling this.

I had a look recently and noted that the core logic that routes log messages - a kind of root logger, if you will - is built around nodejs cluster API (https://nodejs.org/api/cluster.html) which, of course, isn't a thing in the browser.

That had me thinking that the approach of substituting files when building for the browse - which @Download rightly points out - may be adequate but not a very elegant solution. It'd be preferable to change the root logger idea into one which is browser or node, and that root logger requires all the bits it needs.

But naturally this is bigger than it looks and that takes me back to just doing the simplest solution to let webpack work.

Anyone have views and thoughts on best approach?

lamweili:
> Don't solve this in this version ?

This wasn't merged. You might want to see https://github.com/log4js-node/log4js-node/issues/1050 for possible solutions (which sadly no one ever got back on).

aellerton:
Superseded by https://github.com/log4js-node/log4js-node/pull/1374

Issue #1008, opened 2020-05-11
Problem using webpack with fileSync Appender


phaumer:
Hello. I am experiencing the same issue as reported in https://github.com/log4js-node/log4js-node/issues/816 when using the FileSync appender and Webpack.

```
The "path" argument must be of type string. Received type undefined.
```

Could it be that it requires the same solution of adding it to coreAppenders in appenders/index.js ?

phaumer:
Thanks @lauren-li for the fix. My problem is resolved.

PR #1028, opened 2020-06-16
Fix TCP appender with Webpack and Typescript

techmunk:
This is much like the fix for #1008 and #816 
I've also included typescript typings for the TCP appender. I can remove these if you prefer.

There are some other core appenders that are also not in this list, however I'm unsure if this was intentional or by accident so have only included the one I'm currently concerned with and tested for now.


guntram:
Why was this not merged?

swappler-sudo:
What's the status?

peteriman:
@swappler-sudo `Fixed in log4js@6.4.0`

Issue #1060, opened 2021-02-22
%f{depth} filename separator


CarlosGomez-dev:
Hello, it seems I can't get `%f{depth}` to work on my system (Windows 10, Node v14.15), here's a sample codeblock:

```javascript
import log4js from 'log4js';

log4js.configure({
  appenders: {
    out: {
      type: 'console',
      layout: {
        type: 'pattern',
        pattern: '%d %f{1} %m',
      },
    },
  },
  categories: {
    default: { appenders: ['out'], level: 'info', enableCallStack: true },
  },
});

const logger = log4js.getLogger('app');

logger.info('Test log message');
```
Expected output:
`2021-02-22T12:46:30.039 /main.js Test log message`
Actual output:
`2021-02-22T12:46:30.039 file:///C:/Users/Username/Desktop/Projects/Logtest/src/main.js Test log message`

The issue seems to be in `layouts.js` because `path.sep` is ([correctly](https://nodejs.org/api/path.html#path_path_sep)) returning `\`, while filename is `file:///C:/Users/Username/Desktop/Projects/Logtest/src/main.js`, so the filename is never split.

Is this an OS and/or node misconfiguration on my part?

I managed to correct the issue by modifying `layouts.js` split separator parameter, I'll submit a fix PR in case this is a general issue.

peteriman:
I ran your exact codes (saved into test.js) using log4js 6.3.0 on Windows 10 with Node.js 14.
## `node test.js`
> v14.15.0 - `2022-01-04T21:45:59.119 test.js Test log message`
> v14.15.1 - `2022-01-04T21:46:05.225 test.js Test log message`
> v14.15.2 - `2022-01-04T21:46:08.476 test.js Test log message`
> v14.15.3 - `2022-01-04T21:46:13.812 test.js Test log message`
> v14.15.4 - `2022-01-04T21:46:15.480 test.js Test log message`
> v14.15.5 - `2022-01-04T21:46:18.274 test.js Test log message`


Doesn't seem to have any issues.

----

I edited the [layouts.js](https://github.com/log4js-node/log4js-node/blob/master/lib/layouts.js) to have some debug logs:
```javascript
  function fileName(loggingEvent, specifier) {
    let filename = loggingEvent.fileName || '';
    console.log("debug-full-filename:", filename);             // added
    console.log("debug-path.sap:", path.sep);                  // added
    if (specifier) {
      const fileDepth = parseInt(specifier, 10);
[...]

PR #1061, opened 2021-02-22
fix: %f{depth} filename separator

CarlosGomez-dev:
Changed the split parameter to capture both `\` and `/` separators inside array, and then modified the join expression to account for the extra elements in the array. Also removed path import as it's not longer needed.

Related to #1060

### DIFF
--- a/lib/layouts.js
+++ b/lib/layouts.js
@@ -1,7 +1,6 @@
 const dateFormat = require('date-format');
 const os = require('os');
 const util = require('util');
-const path = require('path');
 
 const styles = {
   // styles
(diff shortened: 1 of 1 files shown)


peteriman:
As explained in #1060, I am unable to replicate the issue to warrant this PR.

There shouldn't be any need to hardcode the pattern `\`  and `/` as it is handled by `path.sep`.

> ### path.sep
> Added in: v0.7.9
> [<string>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type)
> Provides the platform-specific path segment separator:
> 
> - `\` on Windows
> - `/` on POSIX
>
> For example, on POSIX:
> 
> ```javascript
> 'foo/bar/baz'.split(path.sep);
> // Returns: ['foo', 'bar', 'baz']
> ```
> On Windows:
> 
> ```javascript
> 'foo\\bar\\baz'.split(path.sep);
> // Returns: ['foo', 'bar', 'baz']
> ```
> On Windows, both the forward slash (`/`) and backward slash (`\`) are accepted as path segment separators; however, the `path` methods only add backward slashes (`\`).
>
> (_src: https://nodejs.org/api/path.html#pathsep_)

peteriman:
This PR has been superseded by #1224.

PR #1072, opened 2021-04-23
fix(types): add defaultLevel parameter declaration to the function ge…

saulzhong:


types/log4js.d.ts
```diff
@@ -270,7 +270,7 @@ export interface Levels {
   FATAL: Level;
   OFF: Level;
   levels: Level[];
-  getLevel(level: string): Level;
+  getLevel(level: string, defaultLevel: Level): Level;
```
peteriman, line 273: Might be better with `getLevel(level: Level | string, defaultLevel: Level): Level;`
https://github.com/log4js-node/log4js-node/blob/f8d46a939279c0ab4efc8bb5f0478c4b0949a4cf/lib/levels.js#L22-L31

peteriman (changes requested):
Can take the opportunity to fix the first parameter as well?

peteriman (approved):

PR #1089, opened 2021-07-14
fix: on newer nodejs versions listening to socket errors are required

jhonatanTeixeira:


peteriman:
From the original code snippet, it is using `net.Socket`.
```javascript
    socket = net.createConnection(config.loggerPort || 5000, config.loggerHost || 'localhost');
```
> ### net.createConnection()
> A factory function, which creates a new [net.Socket](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#class-netsocket), immediately initiates connection with [socket.connect()](https://nodejs.org/dist/latest-v16.x/docs/api/net.html#socketconnect), then returns the `net.Socket` that starts the connection.
>
> _(src: https://nodejs.org/dist/latest-v16.x/docs/api/net.html#netcreateconnection)_
---
Based on `net.Socket` API:
> ### Event: `'error'`
> Added in: v0.1.90
> - [\<Error\>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)
>
> Emitted when an error occurs. The `'close'` event will be called directly following this event.
>
> _(src: https://nodejs.org/dist/latest-v16.x/docs/api/net.html#class-netsocket)_
---
It seems like the behavior did not change and that the `'close'` event will still get called.

@jhonatanTeixeira Would you happen to have any references that newer NodeJS requires listening to socket errors?
Nevertheless, it seems good to flush the buffer on `'error'` before it calls `'close'`.

jhonatanTeixeira:
Just run it on a newer nodejs and try to connect to a tcp destination like logstash for example and you will see it trhows an error. Newer nodejs require all errors thrown inside a promise to be catched, so that cancels the connection

peteriman:
@jhonatanTeixeira You are right.

Error events#
When an error occurs within an EventEmitter instance, the typical action is for an 'error' event to be emitted. These are treated as special cases within Node.js.

If an EventEmitter does not have at least one listener registered for the 'error' event, and an 'error' event is emitted, the error is thrown, a stack trace is printed, and the Node.js process exits.
https://nodejs.org/docs/latest/api/events.html#error-events

(NodeJS event.js)
```javascript
  // If there is no 'error' event listener then throw.
  if (doError) {
    let er;
    if (args.length > 0)
      er = args[0];
    if (er instanceof Error) {
      try {
        const capture = {};
        ErrorCaptureStackTrace(capture, EventEmitter.prototype.emit);
        ObjectDefineProperty(er, kEnhanceStackBeforeInspector, {
          value: FunctionPrototypeBind(enhanceStackTrace, this, er, capture),
          configurable: true
        });
      } catch {}

      // Note: The comments on the `throw` lines are intentional, they show
      // up in Node's output if this results in an unhandled exception.
    }
```

peteriman:
@jhonatanTeixeira I verified. Good find! 

It's due to the way `EventEmitter` works (not `Promises` in this case).

If there is no listener to the `'error'` event, the `EventEmitter` additionally throws the `error`, which in turn becomes `uncaughtException` that terminates the Node.js process.

> # Events
> ### Error events
> When an error occurs within an `EventEmitter` instance, the typical action is for an `'error'` event to be emitted. These are treated as special cases within Node.js.
> 
> 
> ```
> const myEmitter = new MyEmitter();
> myEmitter.emit('error', new Error('whoops!'));
> // Throws and crashes Node.js
> ```
> To guard against crashing the Node.js process the [domain](https://nodejs.org/docs/latest/api/domain.html) module can be used. (Note, however, that the `domain` module is deprecated.)
> 
> As a best practice, listeners should always be added for the `'error'` events.
> 
> ```
> const myEmitter = new MyEmitter();
> myEmitter.on('error', (err) => {
>   console.error('whoops! there was an error');
> });
> myEmitter.emit('error', new Error('whoops!'));
> // Prints: whoops! there was an error
> ```
> 

peteriman (approved):

PR #1095, opened 2021-09-28
Red Color layout fix

qq569513407:
fix problem about red layout


peteriman:
The current red layout (31) works for me though. The new one (91) seems dimmer.

PR #1097, opened 2021-10-04
dateFile error handling fix

4eb0da:
Make the same that "file" appender does

### DIFF
--- a/lib/appenders/dateFile.js
+++ b/lib/appenders/dateFile.js
@@ -28,6 +28,9 @@ function appender(
     options
   );
 
+  logFile.on('error', (err) => {
+    console.error('log4js.dateFileAppender - Writing to file %s, error happened ', filename, err); //eslint-disable-line
+  });
   logFile.on("drain", () => {
     process.emit("log4js:pause", false);
   });
--- a/lib/appenders/dateFile.js
+++ b/lib/appenders/dateFile.js
@@ -22,6 +22,8 @@ function appender(
   // options should work for dateFile as well.
   options.maxSize = options.maxLogSize;
 
+  let alive = true;
+
   const logFile = new streams.DateRollingFileStream(
     filename,
     pattern,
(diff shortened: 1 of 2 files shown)


peteriman (commented):
~~Not so sure if we should implement the `alive` to prevent further writing.~~
~~Perhaps should keep to the current behavior for the time being.~~

**EDIT**: Should be fine since the stream is closed anyway. The `alive` might save some processing (such as removing color).

> The stream is closed when the `'error'` event is emitted unless the [autoDestroy](https://nodejs.org/api/stream.html#new-streamwritableoptions) option was set to `false` when creating the stream.
_(src: https://nodejs.org/api/stream.html#event-error)_

lib/appenders/dateFile.js
```diff
@@ -22,17 +22,26 @@ function appender(
...
 
+  let alive = true;
+
   const logFile = new streams.DateRollingFileStream(
     filename,
     pattern,
     options
   );
 
+  logFile.on('error', (err) => {
```
peteriman, line 33: > If an `EventEmitter` does not have at least one listener registered for the `'error'` event, and an `'error'` event is emitted, the error is thrown, a stack trace is printed, and the Node.js process exits.

Similar to #529 and #1089, the dateFileAppender causes the the Node.js process to exit as there is no listener for its `'error'` event. 
This PR by @4eb0da added the listener to prevent such crashes.

peteriman:

Similar to #529 and #1089, the dateFileAppender causes the the Node.js process to exit as there is no listener for its `'error'` event. 
This PR by @4eb0da added the listener to prevent such crashes.

peteriman:

Similar to #529 and #1089, the dateFileAppender causes the the Node.js process to exit as there is no listener for its `'error'` event. This PR by @4eb0da added the listener to prevent such crashes.

PR #1115, opened 2022-01-07
fix(types): add level parameter declaration to Logger.log function

ZLundqvist:
Minor typedef improvement that better reflects the implementation, and also helps guide TypeScript developers using Logger.log.


peteriman (changes requested):
Might want to remove the extra spacing between `string` and `,`

peteriman (approved):

PR #1141, opened 2022-01-16
Changed default file modes from 0o644 to 0o600 for better security

peteriman:


thernstig:
@peteriman this is a breaking change, as we just noticed this. it breaks things in Kubernetes e.g. when using a sidecar to read logs. (Even though it is configurable the defaults broke). So it probably should have been reserved for 7.X.

peteriman:
You are right. In hindsight, probably should have used 7.x.x instead.
I shall update the changelogs to have a more indicative warning now we are unable to go back in time.

My mistake.

Issue #1161, opened 2022-01-22
Multiprocess appender does not work, master doesn't start


nicojs:
The appender lazy-loading mechanism introduced here ~2 years ago broke the multiprocess appender.

Commit here: https://github.com/log4js-node/log4js-node/commit/3cb7f52fce3d4721685649d33999f36443477eac

PR #1170, opened 2022-01-25
test(multiprocess): add e2e test

nicojs:
Prevent regression for the multiprocess appender.

Regression like #1161.


nicojs:
I noticed that an actual e2e test was missing for the multiprocess appender. This e2e test would prevent regressions as we've had in the past.

PR #1197, opened 2022-03-03
Removed redundant logic in multiprocessAppender

peteriman:


lib/appenders/multiprocess.js
```diff
@@ -32,10 +32,8 @@ function logServer(config, actualAppender, levels) {
     let logMessage = '';
 
     function logTheMessage(msg) {
-      if (logMessage.length > 0) {
```
peteriman, line 35: `logMessage.length` will never be <= 0 because it was checked in Line 45.
https://github.com/log4js-node/log4js-node/blob/3dfa03e36bd81e36fc2860245d26789dc6d63528/lib/appenders/multiprocess.js#L41-L52

PR #1198, opened 2022-03-04
Removed redundant logic in tcp-serverAppender

peteriman:


lib/appenders/tcp-server.js
```diff
@@ -7,46 +7,41 @@ const DELIMITER = '__LOG4JS__';
 
 exports.configure = (config) => {
   debug('configure called with ', config);
-  // dummy shutdown if we're not master
-  let shutdown = (cb) => { cb(); };
 
-  clustering.onlyOnMaster(() => {
```
peteriman, line 13: No need for `clustering.onlyOnMaster()` as it will be checked by its caller in `appenders/index.js` Line 87.
https://github.com/log4js-node/log4js-node/blob/b02c07c34af9560e2180617c8d3fb6c4fed331d3/lib/appenders/index.js#L87-L95

peteriman, line 13: No need for `clustering.onlyOnMaster()` as it is already checked by its caller in `appenders/index.js` Line 87.
https://github.com/log4js-node/log4js-node/blob/b02c07c34af9560e2180617c8d3fb6c4fed331d3/lib/appenders/index.js#L87-L95

PR #1267, opened 2022-06-09
Make Appender Type extensible from other modules and the user

ZachHaber:
Appender was a discriminated type union, which unfortunately made it so that it wasn't able to be modified. Now you can update the `Appenders` interface with new values to enable modification.

```ts
declare module './log4js' {
  interface Appenders {
    StorageTestAppender: {
      type: 'storageTest';
      storageMedium: 'dvd' | 'usb' | 'hdd';
    };
  }
}
```
------
Most of the changes in this PR are prettier being run, as I also set up the editorconfig to apply properly to typescript files as well.

The gitattributes was also missing typescript files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[*.js]
+[*.{js,ts,[cm]js,[cm]ts}]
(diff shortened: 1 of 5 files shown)


peteriman (approved):
@ZachHaber Thanks! 👍

PR #1271, opened 2022-06-20
Run prettier and require it

ZachHaber:
@peteriman, I ran prettier accross the entire project, and like the logfaces-http, it now is part of the pretest config.


peteriman:
It broke the Windows tests?

ZachHaber:
> Hmm, it broke the Windows tests?
> 
> The created test files should have been deleted on test teardown I believe.

That one particular test file has been appearing on my computer every time I ran the test. Even on the previous branch I used. I never opened the configuration-validation-test, but it started appearing one day, and then it always appeared after every time I ran the test afterwards.

ZachHaber:
It looks like the windows node 16 test should work faster after we run an `npm install` or update locally to let the `package-lock.json` file update to version 2.

peteriman:
I made some changes to the commits:
- merged the two commits for `.gitattributes` together
- added a commit (https://github.com/log4js-node/log4js-node/pull/1271/commits/e378b597a6ae4b90638371c8eb4e67aa88a8f533) to refactor eslint comments so prettier can fix properly
- removed the commit that added `test/tap/freeze-date-file-test` to `.gitignore`

PR #1316, opened 2022-08-27
FEAT: patternLayout function name, class name and function alias

l0ner:
Added to patternLayout the following fields:

- `%M` - function name of the caller issuing the logging request
- `%C` - class name of the caller issuing the logging request
- `%A` - function alias of the caller issuing the logging request

The first two come from log4j, and use the same field specifiers.

When called from method `bar` of a class `Foo`:

- `%M` will be replaced with `bar`
- `%C` will be replaced with `Foo`
- `%A` will be empty

When called from function foo:

- `%M` will be replaced with foo
- `%C` will be empty
- `%A` will be empty

`%A` will be non empty only if the call stack parsed to obtain the values contains string `[as foo]`.

This is the same behaviour as log4j (except for the `%A`).

Advise: would it make sense to also include something like `%F` for obtaining the full, unsplitted name?

------

Added table for clarity:
|Error Stacktrace|%M<br>function name|%C<br>class name|%A<br>alias name|%F<br>caller name|
|--|--|--|--|--|
|\<empty\>| | | | |
|`foo`|`foo`| | |`foo`|
|`foo [as bar]`|`foo`||`bar`|`foo [as bar]`|
|`Foo.bar`|`bar`|`Foo`||`Foo.bar`|
|`Foo.bar [as baz]`|`bar`|`Foo`|`baz`|`Foo.bar [as baz]`|


lamweili:
> Advise: would it make sense to also include something like `%F` for obtaining the full, unsplitted name?

Can you provide a sample for my understanding? 🤗

l0ner:
> > Advise: would it make sense to also include something like `%F` for obtaining the full, unsplitted name?
> 
> Can you provide a sample for my understanding? 🤗

The names come from splitting an input string, that comes from the call stack. This string can be anything between:

- empty (when the caller is in a file scope or an anonymous function in a file scope)
- `foo` - caller is a function `foo`
- `foo [as bar]` - caller is a function `foo` imported as `bar`
- `Foo.bar` - caller is method `bar` of a class `Foo`
- `Foo.bar [as baz]` - caller is method `bar` of a class `Foo` imported as `baz`

Somehing like `%F` (function name) would be replaced by one of this possibile strings. Could prove useful when user wants fully qualified caller name, insted of wanting only function/class/alias name.

lamweili:
I see. So what would be `%F` (instead of concatenation manually)?
Is there a priority from your list?

lamweili:
I see. So `%F` would be the entire group of the regexp?
I think that's a good idea 👍, so there is no need to do concatenation manually.

You must be careful to ensure the test cases have good coverage for the various permutations/combinations for the match.

l0ner:
I have added the `%F` field containing fully qualified caller name.

Test should be good, but please review, I've tried my best to add all possible permutations.

lamweili:
I think you forgot to update the `layouts.md` for `%F`

lamweili:
I think you forgot to update the `layouts.md` for `%F`.

lamweili (changes requested):
Hi, I made a review and have some comments.
Could you please address them? 🙏

lib/logger.js
```diff
@@ -15,12 +15,30 @@ function defaultParseCallStack(data, skipIdx = 4) {
...
     /* istanbul ignore else: failsafe */
     if (lineMatch && lineMatch.length === 6) {
+      // extract class, function and alias names
+      let className = '';
+      let functionName = '';
+      let functionAlias = '';
+      if (lineMatch[1] && lineMatch[1] !== '') {
+        // WARN: this will unset alias if alias is not present.
+        [functionName, functionAlias] = lineMatch[1]
+          .replace(/[[\]]/g, '')
```
lamweili, line 25: I see that this removes `[` and `]`.
Can you explain when this is needed?

test/tap/LoggingEvent-test.js
```diff
@@ -97,5 +97,176 @@ test('LoggingEvent', (batch) => {
...
+    t.equal(event.fileName, fileName);
+    t.equal(event.lineNumber, lineNumber);
+    t.equal(event.columnNumber, columnNumber);
+    t.equal(event.callStack, callStack);
+    t.equal(event.className, className);
+    t.equal(event.callerName, callerName);
+    t.equal(event.functionAlias, '');
+    t.end();
+  });
+
```
lamweili, line 231: I think there is no need for these test cases as it's a straightforward variable assignment.
The `Should contain class, method and alias names` test case would have covered them.

```suggestion
```

docs/layouts.md
```diff
@@ -143,6 +143,9 @@ Fields can be any of:
 - `%l` line number (requires `enableCallStack: true` on the category, see [configuration object](api.md))
 - `%o` column postion (requires `enableCallStack: true` on the category, see [configuration object](api.md))
 - `%s` call stack (requires `enableCallStack: true` on the category, see [configuration object](api.md))
+- `%M` function or method name (requires `enableCallStack: true` on the category, see [configuration object](api.md))
+- `%C` class name (requires `enableCallStack: true` on the category, see [configuration object](api.md))
+- `%A` function or method name alias (requires `enableCallStack: true` on the category, see [configuration object](api.md))
```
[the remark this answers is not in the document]
l0ner: Added in commit [397b06e](https://github.com/log4js-node/log4js-node/pull/1316/commits/397b06e15f5a6d485cafdbb141ab80b916eb9004)

lib/layouts.js
```diff
@@ -109,6 +109,9 @@ function dummyLayout(loggingEvent) {
  *  - %l line number
  *  - %o column postion
  *  - %s call stack
+ *  - %M method or function name
+ *  - %C class name
+ *  - %A method or function name
```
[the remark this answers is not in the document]
l0ner: Added in commit [397b06e](https://github.com/log4js-node/log4js-node/pull/1316/commits/397b06e15f5a6d485cafdbb141ab80b916eb9004)

lamweili (changes requested):

test/tap/logger-test.js
```diff
@@ -253,6 +253,51 @@ test('../../lib/logger', (batch) => {
     t.end();
   });
 
+  batch.test('parseCallStack names extraction', (t) => {
```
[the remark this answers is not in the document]
l0ner: Added in commit [4360da0](https://github.com/log4js-node/log4js-node/pull/1316/commits/4360da010fffba1c1fcec8b9412df0da6581d8dd)

test/tap/LoggingEvent-test.js
```diff
@@ -97,5 +97,176 @@ test('LoggingEvent', (batch) => {
...
+    t.equal(event2.fileName, undefined);
+    t.equal(event2.lineNumber, undefined);
+    t.equal(event2.columnNumber, undefined);
+    t.equal(event2.callStack, undefined);
+    t.equal(event2.className, undefined);
+    t.equal(event2.functionAlias, undefined);
+    t.equal(event2.callerName, undefined);
+    t.end();
+  });
+
```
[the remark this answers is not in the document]
l0ner: Fixed in commit [4360da0](https://github.com/log4js-node/log4js-node/pull/1316/commits/4360da010fffba1c1fcec8b9412df0da6581d8dd)

test/tap/LoggingEvent-test.js
l0ner: Removed all three in commit [5f988bd](https://github.com/log4js-node/log4js-node/pull/1316/commits/5f988bd097b4b5218fd564a2a1cd0466c5501f0f)

test/tap/LoggingEvent-test.js
```diff
@@ -94,6 +106,9 @@ test('LoggingEvent', (batch) => {
     t.equal(event2.lineNumber, undefined);
     t.equal(event2.columnNumber, undefined);
     t.equal(event2.callStack, undefined);
+    t.equal(event2.className, undefined);
```
lamweili, line 109: Forgot to include `t.equal(event2.functionName, undefined);`?

lamweili (changes requested):

test/tap/LoggingEvent-test.js
l0ner: Yeah, noticed it myself too. Fixed in [6ea3e43](https://github.com/log4js-node/log4js-node/pull/1316/commits/6ea3e43c62f2d075a335b87967e3b1c8a1b28067)

lamweili:
Anyway, I have also edited your original post with a table for clarity.
Feel free to correct it if it's wrong.

l0ner:
> Anyway, I have also edited your original post with a table for clarity. Feel free to correct it if it's wrong.

It's good. No need for changes.

test/tap/LoggingEvent-test.js
```diff
@@ -106,6 +106,7 @@ test('LoggingEvent', (batch) => {
     t.equal(event2.lineNumber, undefined);
     t.equal(event2.columnNumber, undefined);
     t.equal(event2.callStack, undefined);
+    t.equal(event.functionName, undefined);
```
lamweili, line 109: Tests didn't pass due to typo.
```suggestion
    t.equal(event2.functionName, undefined);
```

lamweili (changes requested):

test/tap/LoggingEvent-test.js
l0ner: Fixed.

lamweili (approved):

l0ner:
Great to see this merged! Can't wait the see this available on npm so I can use it in my work.

vineetmyhq:
When are we expecting a release for this? 
Really need this feature when using logger in a wrapper, where can't track who generated the log.

lamweili:
Target by end of the week. 🤗

lamweili:
Released in `log4js@6.7.0`.

PR #1374, opened 2023-03-07
Idempotent logging on browser

aellerton:
Certain modules do not exist in a browser environment, e.g. cluster, file/os/path, etc.

In that environment, file-based appenders are effectively disabled by this commit, without modifying the source directly.

The `clustering.js` file is still required, but this commit essentially makes it a no-op, again without modifying the source directly.

This addresses #968 and supersedes PR 1023 https://github.com/log4js-node/log4js-node/pull/1023

### DIFF
--- a/lib/clusteringBrowser.js
+++ b/lib/clusteringBrowser.js
@@ -1,0 +1,19 @@
+/* istanbul ignore file */
+// This is used in browsers only and is designed to allow the rest of
+// log4js to continue as if `clustering.js` is in use.
+const isMaster = () => true;
+
+const listeners = [];
+
+const sendToListeners = (logEvent) => {
+  listeners.forEach((l) => l(logEvent));
+};
+
+module.exports = {
+  onlyOnMaster: (fn, notMaster) => (isMaster() ? fn() : notMaster),
+  isMaster,
+  send: sendToListeners,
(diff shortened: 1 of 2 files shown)


lamweili:
Ah, some misc prettier errors on CI: `npm run prettier:fix`

lamweili:
Now you are missing some tests for the test coverage and you should be good to go!

lamweili:
Now you are missing some tests for the test coverage and you should be good to go!
May want to use `/* istanbul ignore file */` if it doesn't fit in nicely.

lib/clusteringBrowser.js
```diff
@@ -0,0 +1,18 @@
+// This is used in browsers only and is designed to allow the rest of
```
lamweili, line 1: This might work since we can't really test it, can we?

```suggestion
/* istanbul ignore file */
// This is used in browsers only and is designed to allow the rest of
```

aellerton: That's a great suggestion, thank you!

lamweili:
Fixes #968

aellerton:
Thank you!

lamweili:
> Thank you!

I should thank you instead! :)

PR #1375, opened 2023-03-07
docs: added that `log4js.getLogger()` may call `log4js.configure()`

lamweili:


docs/api.md
```diff
@@ -32,6 +32,8 @@ Properties:
 
 ## Loggers - `log4js.getLogger([category])`
 
+To support the minimalist usage, this function will implicitly call `log4js.configure()` with the [default configurations](https://github.com/log4js-node/log4js-node/blob/faebee0e8235cf00227ca88642eeff3471fc407c/lib/log4js.js#L158-L163) if it hasn't been configured before.
```
thernstig, line 35: This has a chance to quickly become outdated in case you change the default configuration.

PR #1402, opened 2023-11-03
docs(file): add `removeColor` option of file appender

mkobayashime:
Add `removeColor` option of file appender (merged in #1001), which is missing in the docs.


lamweili:
Thanks!

